PR #20, opened 2022-03-10
An 867 magic eden

jhuhnke:


models/silver/silver__nfts_magic_eden_v1.sql
```diff
@@ -0,0 +1,47 @@
...
+      index, 
+      program_id, 
+      instruction :accounts[0] :: STRING AS purchaser,  
+       COALESCE (inner_instruction:instructions[0]:parsed:info:lamports/POW(10,9), 0) as price_0, 
+       COALESCE (inner_instruction:instructions[1]:parsed:info:lamports/POW(10,9), 0) as price_1,
+       COALESCE (inner_instruction:instructions[2]:parsed:info:lamports/POW(10,9), 0) as price_2, 
+       COALESCE (inner_instruction:instructions[3]:parsed:info:lamports/POW(10,9), 0) as price_3, 
+      COALESCE (inner_instruction:instructions[4]:parsed:info:lamports/POW(10,9), 0) as price_4, 
+      COALESCE (inner_instruction:instructions[5]:parsed:info:lamports/POW(10,9), 0) as price_5,
+      (price_0 + price_1 + price_2 + price_3 + price_4 + price_5) AS sale_amount,
```
desmond-hui, line 22: The current methodology will not scale, it is capped at 5 amounts.  If there are 6 or more amounts, the sales price is not correct.  Example `2QY63qHsgW9A9NmBQpceCAe4kLSMPTjyroo5J6X2pkZ7zws2UmzRnoYSC1DmCbRgHRdM3gzh7wmXp1p5ddP4R12X`

This should be done by flattening the `inner_instruction:instructions` so that you get an a row per `inner_instruction`.  Use that to pull out the solana amount within each `inner_instruction` and then group the records back together and sum the amounts.

models/silver/silver__nfts_magic_eden_v2.sql
```diff
@@ -0,0 +1,42 @@
...
+    p.mint AS NFT, 
+    inner_instruction :instructions[0] :parsed :info :lamports / POW(10,9) :: INTEGER AS sale_amount, 
+    e.ingested_at
+FROM txs t
+
+INNER JOIN {{ ref('silver__events') }} e
+ON t.tx_id = e.tx_id 
+
+INNER JOIN {{ ref('silver___post_token_balances') }} p
+ON e.tx_id = p.tx_id AND e.index = p.index
```
desmond-hui, line 39: Same comments as for V1 instruction.  I think you can find the mint from the inner_instructions for a v2 sale

models/silver/silver__nfts_magic_eden_v2.sql
```diff
@@ -0,0 +1,42 @@
...
+    e.ingested_at
+FROM txs t
+
+INNER JOIN {{ ref('silver__events') }} e
+ON t.tx_id = e.tx_id 
+
+INNER JOIN {{ ref('silver___post_token_balances') }} p
+ON e.tx_id = p.tx_id AND e.index = p.index 
+
+WHERE t.ct = 3
```
desmond-hui, line 41: I think there are NFT sales w/ < 3 events.

https://solscan.io/tx/4P2rPf6swFa27aBqaRxLso99ctB72rxVPGoA5yt7PREUzDf1CGpnAkMr5t2Lfd6DG3jJoDNT3L8zy2CrcjyVpFJS

^ this one looks like a sale to me, but it doesnt have the instruction that gives the sale total

models/silver/silver__nfts_magic_eden_v2.sql
```diff
@@ -0,0 +1,42 @@
...
+        tx_id, 
+        count(tx_id) AS ct
+    FROM {{ ref('silver__events') }}
+    WHERE program_id = 'M2mx93ekt1fmXSVkTrUL9xVFHkmME8HTUi5Cyc5aF7K' -- Magic Eden V2 Program ID
+    
+    {% if is_incremental() %}
+      AND ingested_at >= getdate() - interval '2 days'
+    {% endif %}
+    
+    GROUP BY tx_id
```
desmond-hui, line 19: If you need to do a count filter, you can do it here with a `having` clause.

`GROUP BY tx_id
HAVING count(tx_id) = 3`

desmond-hui (changes requested):

models/silver/silver__nfts_magic_eden_v1.sql
```diff
@@ -0,0 +1,47 @@
...
+    t.index, 
+    t.program_id, 
+    purchaser, 
+    p.mint AS NFT, 
+    sale_amount, 
+    t.ingested_at
+FROM txs t
+
+INNER JOIN "SOLANA_DEV"."SILVER"."_POST_TOKEN_BALANCES" p
+ON t.tx_id = p.tx_id AND t.index = p.index
```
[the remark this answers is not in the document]
jhuhnke: Not even sure this is necessary. There is a mint field in the inner instructions - we can just pull that.

models/silver/silver__nfts_magic_eden_v2.sql
```diff
@@ -0,0 +1,123 @@
...
+
+  INNER JOIN {{ ref('silver__events') }} e
+  ON t.tx_id = e.tx_id
+
+  WHERE e.index = t.ct - 1
+
+  {% if is_incremental() %}
+      AND ingested_at::date >= getdate() - interval '2 days'
+  {% endif %}
+),
```
desmond-hui, line 43: This CTE isnt necessary.  You can get the max event index for the transaction in the above CTE since you are already grouping by transaction and select from the events table

`max(index) as last_index`

models/silver/silver__nfts_magic_eden_v2.sql
```diff
@@ -0,0 +1,123 @@
...
+      program_id,
+      amount,  
+      ingested_at
+    FROM amounts_agg a
+    
+    INNER JOIN max_index_remove m
+    ON m.tx_id = a.tx_id
+  
+    WHERE index <> m.max_index   
+),
```
desmond-hui, line 96: I think this can be reduced to a single CTE where we filter to the inner_instructions we want per transaction.  Again you can use `max()` to figure out the max inner instruction index per transaction so that you can use it to ignore that record down the line.  See example below, you don't need a group by if you use `over (partition by ...)`

`max(i.index) over (partition by e.tx_id, e.index, NFT) as max_inner_index`

desmond-hui (changes requested):

models/silver/silver__nfts_magic_eden_v1.sql
```diff
@@ -0,0 +1,56 @@
...
+    p.mint AS NFT, 
+    t.owner AS purchaser, 
+    t.ingested_at
+FROM txs t
+
+INNER JOIN sales_amount s
+ON s.tx_id = t.tx_id
+
+INNER JOIN "SOLANA_DEV"."SILVER"."_POST_TOKEN_BALANCES" p
+ON p.tx_id = t.tx_id AND p.account = t.account
```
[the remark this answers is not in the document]
desmond-hui: table reference needs to be parameterized

models/silver/silver__nfts_magic_eden_v1.sql
```diff
@@ -0,0 +1,56 @@
...
+        block_timestamp, 
+        block_id, 
+        tx_id, 
+        program_id,  
+        i.value :parsed :info :lamports / POW(10,9) AS amount, 
+        i.value :parsed :info :account :: STRING AS account, 
+        i.value :parsed :info :newAuthority :: STRING as owner, 
+        inner_instruction, 
+        ingested_at
+    FROM "SOLANA_DEV"."SILVER"."EVENTS",
```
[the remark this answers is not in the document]
jhuhnke: I missed this... whoops

models/silver/silver__nfts_magic_eden_v1.yml
```diff
@@ -0,0 +1,41 @@
+version: 2
+models:
+  - name: silver__nfts_magic_eden_v1
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - TX_ID
+            - NFT
```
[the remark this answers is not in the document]
jhuhnke: This needs to be the unique key due to bundled sales of NFTs. https://solscan.io/tx/31jptwqYr8EPqKBhrS4Ah1eyZhqV3mTh8YAMbAtXo5SgVNpYRCs6foEGkT475aeP2tbfytpLenJsUU66AyvHo2Ed

Going to represent the price differently

models/silver/silver__nfts_magic_eden_v2.sql
jhuhnke: Found an easier way to pull the price rather than sum everything - killed off all these CTEs and used the easier way

nickpayiatis (approved):

PR #26, opened 2022-03-18
An 882 candy machine

jhuhnke:


models/silver/silver__nfts_candy_machine.sql
```diff
@@ -0,0 +1,83 @@
...
+)       
+
+SELECT 
+    block_timestamp, 
+    block_id, 
+    t.tx_id, 
+    program_id,
+    succeeded, 
+    max(wallet) AS purchaser,  
+    max(sales_amount / POW(10, COALESCE(p.decimal, 9))) AS sales_amount,
```
desmond-hui, line 71: Are minting NFTs actually considered a sale?  If not we should change the names here

models/silver/silver__nfts_candy_machine.sql
```diff
@@ -0,0 +1,83 @@
...
+
+   WHERE program_id = 'cndyAnrLdpjq1Ssp1z8xxDsB8dxe7u4HL5Nxi2K5WXZ' 
+   AND e.index > 0
+   AND (authority IS NOT NULL 
+   OR NFT IS NOT NULL
+   OR i.index = 0) 
+   AND t.succeeded = TRUE
+
+   {% if is_incremental() %}
+    AND e.ingested_at :: DATE >= current_date - 2
```
desmond-hui, line 41: Add incremental for `silver__transactions`

models/silver/silver__nfts_candy_machine.yml
```diff
@@ -0,0 +1,51 @@
...
+        tests:
+          - not_null
+      - name: TX_ID
+        description: "{{ doc('tx_id') }}"
+        tests:
+          - not_null
+      - name: PROGRAM_ID 
+        description: "{{ doc('program_id') }}"
+        tests: 
+          - not_null
```
desmond-hui, line 28: Add description for `succeeded` or exclude it from the output if we only have successful mints

desmond-hui (changes requested):
I think we should change the name to `silver__nft_candy_machine_v1.sql` to match existing convention.

PR #29, opened 2022-03-21
adds silver NFT model for Solanart marketplace

jhuhnke:


models/silver/silver__nft_solanart.sql
```diff
@@ -0,0 +1,67 @@
...
+    s.succeeded, 
+    s.program_id, 
+    max(p.mint) AS mint, 
+    s.purchaser, 
+    SUM(s.amount) / POW(10,9) AS sales_amount, 
+    s.ingested_at
+FROM sales_inner_instructions s
+
+INNER JOIN post_token_balances p
+ON s.tx_id = p.tx_id
```
desmond-hui, line 63: I think we can join this using `tx_id and account` so you don't need to do a `max(mint)`.  Looks like Solanart gives us the buyer token account directly in instructions so you don't need to search for it.  You can just grab it using `e.instruction :accounts [1] :: STRING` then join that to the post token balances account.

Example:  https://solscan.io/tx/4eupCPLhzNPsPndinUZ1bfFLNoGzYxFqjizCwvJJ2T4bPa5oPYsHPupD735tS7nhJ2uRQ9hBWcxjmt8NLgk1VkoA

models/silver/silver__nft_solanart.sql
```diff
@@ -0,0 +1,67 @@
+{{ config(
+  materialized = 'incremental',
+  unique_key = "CONCAT_WS('-', tx_id, mint)",
+  incremental_strategy = 'delete+insert',
+  cluster_by = ['block_timestamp::DATE'],
+) }}
```
desmond-hui, line 6: Modify name of model to `nft_sales_solanart`

desmond-hui (changes requested):

desmond-hui (approved):

PR #32, opened 2022-03-23
adds core dim nft metadata table & updates NFT sales

jhuhnke:


models/core/core__dim_nft_metadata.yml
```diff
@@ -0,0 +1,62 @@
+version: 2
+models:
+  - name: core__dim_nft_metadata
```
desmond-hui, line 3: Change these tests to match the other core views.  We don't really need to check for NOT NULLS when we are doing a simple select w/o any logic.  The silver tests already cover this.  Instead we can check that the expected columns are in the view

desmond-hui (changes requested):

desmond-hui (approved):

PR #40, opened 2022-04-18
Removes metadata with null mint at silver level

jhuhnke:


eric-laurello (changes requested):
Not sure how your change is causing this failure, but on lines 24-29 that subquery is invalid. _inserted_timestamp doesnt exist on the silver.nft_metadata table. 

Does this run as a full every time?

eric-laurello:
I'm not seeing it in the table, but the error is actually slightly different:
 Subquery containing correlated aggregate function [MAX(PROD_NFT_METADATA_UPLOADS_1828572827._INSERTED_TIMESTAMP)] can only appear in having or select clause

eric-laurello (approved):

PR #41, opened 2022-04-26
add nfts related to marindate governance

desmond-hui:
- Add marinade NFT mints related to MNDE locks for governance
- Price for these mints will be 0 because the value is in the MNDE lock which can be increased over-time.  This MNDE value will be tracked in a separate table related to governance


forgxyz (approved):
LGTM!

PR #58, opened 2022-06-15
fix issue with granting ownership for funcs with params

desmond-hui:
- fix issue with granting proper perms for cloned functions that have parameters


amasucci13 (approved):
Approved

PR #59, opened 2022-06-15
incorporate additional nft mint patterns and programs

desmond-hui:
- Fix to records having incorrect value for MINT
- Fix issue with some mint transactions not appearing in the table


amasucci13 (approved):
Approved

PR #69, opened 2022-07-14
An 1705 ez staking v2 jessica

jhuhnke:


models/silver/silver___historical_staking_lp_actions.sql
```diff
@@ -0,0 +1,94 @@
...
+        index, 
+        ii.value :parsed :type :: STRING AS event_type, 
+        ii.value :programId :: STRING AS program_id,
+        ii.value AS instruction
+    FROM 
+        {{ source(
+            'solana_external', 
+            'txs_api') 
+        }}, 
+    TABLE(FLATTEN (data :result :meta :innerInstructions[0] :instructions)) ii
```
desmond-hui, line 36: This needs to flatten out the list of innerInstructions and then flatten out the instructions to get all innerinstructions that have the staking programid

Example of tx that is not in the historical because the staking instructions are in the 2nd element of innerinstructions
`ibm7xuok39VvjF4BfTSF4Lz8EYZMp9ZJhehsi9P7z2q41u9qRujQXoBdjvdABfFckiBvrYvXJxg8s1TzU4Gcogj`

Sample code for grabbing all the staking instructions for that tx

`select tx_id, 
        CONCAT(
            b.value:index,
            '.',
            c.index
        ) AS INDEX,
        c.value :parsed :type :: STRING AS event_type, 
        c.value :programId :: STRING AS program_id,
        c.value AS instruction,
        b.value
from solana_dev.bronze.txs_api a,
table(flatten(data :result :meta :innerInstructions)) b,
table(flatten(b.value:instructions)) c
where tx_id = 'ibm7xuok39VvjF4BfTSF4Lz8EYZMp9ZJhehsi9P7z2q41u9qRujQXoBdjvdABfFckiBvrYvXJxg8s1TzU4Gcogj'
and program_id = 'Stake11111111111111111111111111111111111111';`

desmond-hui (changes requested):

desmond-hui (approved):

PR #76, opened 2022-08-01
An 1719 realms

jhuhnke:
Vote options are found in the transaction log. Yes and no votes have two separate log structures. 

There is a default program ID for realms that starts with GovER... However, if you write your own gov contract your space will have a unique program ID. There are multiple important spaces, such as Mango Markets, Serum, and Orca that utilize their own program ID. Labels for all the governance spaces' program IDs were uploaded from the Solana labs github to the labels table and pulled in via a CTE to this model to ID votes that occur on Realms. 

The final "WHERE" clause is necessary to correctly represent three transactions in which two vote accounts were created and two votes occurred within the same transaction. These "double" votes were done by the same voter on the same transaction. The following tx_ids are the transactions in question: 'jGD6MefUkncsxnaXVznVxm5UCfzw6ENoecfC2EBvo8JLpXFUGMW2tRcFJp3v1irL4b8ohoBv5C5oS23xcsYW1Tt', '2mydbkQycUMNwSmjuwbunagB3uNpyJQ3eq7gVQRYfEc5MC6opkouf5GRiZFaXg2FNinVb4bPhVgsAJK4e7Zx9KNG', and 'ctzYATbWK98U6v3RNcpxWBDURe8bAGMptaLtHKRuTVr9xPNFE7bqdhQx6yLLGf2StQQ5hj461am2u28fhY2GGAP'

### DIFF
--- a/models/silver/silver__proposal_votes_realms.sql
+++ b/models/silver/silver__proposal_votes_realms.sql
@@ -45,15 +45,15 @@ SELECT
     realms_id, 
     proposal, 
     voter, 
-    split_part(split_part(split_part(split_part(l.value :: STRING, '{', 3), '}', 1), ',', 1), ':', 2) :: INTEGER AS vote_choice, 
-    split_part(split_part(split_part(split_part(l.value :: STRING, '{', 3), '}', 1), ',', 2), ':', 2) :: INTEGER AS vote_weight, 
+    split_part(split_part(split_part(split_part(l.value :: STRING, '{', 3), '}', 1), ',', 1), ':', 2) AS vote_choice, 
+    split_part(split_part(split_part(split_part(l.value :: STRING, '{', 3), '}', 1), ',', 2), ':', 2) AS vote_weight, 
     t._inserted_timestamp
-FROM {{ ref('silver__transactions') }} t
+FROM vote_txs v 
     
-LEFT OUTER JOIN TABLE(FLATTEN(t.log_messages)) l
-    
-LEFT OUTER JOIN vote_txs v
+LEFT OUTER JOIN {{ ref('silver__transactions') }} t
 ON v.tx_id = t.tx_id
+
+LEFT OUTER JOIN TABLE(FLATTEN(t.log_messages)) l
     
 WHERE 
     v.block_timestamp :: date = '2022-07-19' 
--- a/models/silver/silver__proposal_votes_realms.sql
+++ b/models/silver/silver__proposal_votes_realms.sql
@@ -26,7 +26,7 @@ vote_txs AS (
         program_id, 
         instruction :accounts[0] :: STRING AS realms_id, 
         instruction :accounts[2] :: STRING AS proposal, 
-        instruction :accounts[2] :: STRING AS voter, 
+        instruction :accounts[5] :: STRING AS voter, 
         instruction :accounts[6] :: STRING AS vote_account, 
         _inserted_timestamp
     FROM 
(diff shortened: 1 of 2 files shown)


models/silver/silver__proposal_votes_realms.sql
```diff
@@ -0,0 +1,112 @@
...
+INNER JOIN TABLE(FLATTEN(t.log_messages)) l
+    
+WHERE 
+    (l.value :: STRING LIKE 'Program log: GOVERNANCE-INSTRUCTION: CastVote %' 
+    AND v.index = 0 
+    AND _index <= 6)
+    OR (
+    l.value :: STRING LIKE 'Program log: GOVERNANCE-INSTRUCTION: CastVote %'
+    AND v.index = 1 
+    AND _index >= 7)
```
desmond-hui, line 102: These index conditions will cause the model to miss voting transactions.  Here are some examples that I don't see in the current model output but are realm votes

https://solscan.io/tx/X9QQH9YUNYRR1nMnozDWLD37xF5Wr9r1rxJ5dp1uYRSasZZ4EDeXTsfKL9JC65XgxotmJ7HAj1P7218fk84YdN9

https://solscan.io/tx/5nWKYz2p4WiM4DvXVwLhrVAJfHtKuKD6XHET5ti2UZd9Ea9eGD4hUXTeTazsu9rS6Gj9uehEDb3cMzduBiNALPEW

Instead of hardcoding the index values, you can map the log messages back to the individual instructions that the log message is for.  I think in this case we are concerned with mapping back to the event(instruction) level index only.  Check out `models/silver/silver__gauges_creates_marinade.sql` to see how this can be done.  Look at CTEs `b, C, tx_logs`

models/silver/silver__proposal_votes_realms.sql
```diff
@@ -0,0 +1,112 @@
...
+    SELECT
+        block_timestamp, 
+        block_id, 
+        tx_id,
+        succeeded, 
+        e.index, 
+        program_id, 
+        instruction :accounts[0] :: STRING AS realms_id, 
+        instruction :accounts[2] :: STRING AS proposal, 
+        i.value :parsed :info :source :: STRING AS voter,
```
desmond-hui, line 30: Is the voter always account 5 on the instructions level?  If so, we can remove the need to flatten out the inner instructions

desmond-hui (changes requested):

models/silver/silver__proposal_votes_realms.sql
jhuhnke: Looks like this is correct. Removing flatten

models/silver/silver__proposal_votes_realms.sql
```diff
@@ -0,0 +1,166 @@
...
+    SELECT
+        block_timestamp, 
+        block_id, 
+        tx_id,
+        succeeded, 
+        e.index, 
+        program_id, 
+        instruction :accounts[0] :: STRING AS realms_id, 
+        instruction :accounts[2] :: STRING AS proposal, 
+        instruction :accounts[2] :: STRING AS voter,
```
desmond-hui, line 29: Wrong account index for voter

desmond-hui (changes requested):

desmond-hui (approved):

models/silver/silver__proposal_votes_realms.sql
```diff
@@ -0,0 +1,166 @@
...
+        ON t.tx_id = v.tx_id
+        LEFT OUTER JOIN TABLE(FLATTEN(t.log_messages)) l
+
+{% if is_incremental() %}
+AND _inserted_timestamp >= (
+    SELECT
+        MAX(_inserted_timestamp)
+    FROM
+        {{ this }}
+)
```
desmond-hui, line 85: This incremental portion should be `WHERE` instead of `AND`

desmond-hui (changes requested):

desmond-hui (approved):

PR #77, opened 2022-08-06
realms to gold

jhuhnke:


models/core/core__fact_proposal_votes.sql
```diff
@@ -11,6 +11,31 @@ SELECT
...
+    'NULL' AS vote_rank, 
+    'NULL' AS vote_weight
+
+FROM
+    {{ ref('silver__proposal_votes_marinade') }}
+
+UNION ALL 
+
+SELECT 
+    'realms' as program_name,
```
desmond-hui, line 26: This is generally for the "label" associated w/ the specific solana program that the proposal vote is for.. i.e should be Mango, Solend, Marinade, etc.  If we don't know the label then maybe we should use the program_id instead or consider putting realm votes into it's own table?  We could also add a new column to indicate whether something is realms or tribeca

desmond-hui (changes requested):

desmond-hui (approved):

PR #83, opened 2022-08-15
add macro to be called on dbt cloud

desmond-hui:
- Remove dev refresh snowflake task
- Add dbt macro for calling the dev refresh stored procedure


amasucci13 (approved):
Approved

PR #119, opened 2022-10-06
An 1903/signers

jhuhnke:


models/silver/silver__daily_signers.sql
```diff
@@ -0,0 +1,179 @@
...
+                LEAST(COALESCE(MAX(block_id), 105368)+1,153013616)
+            FROM
+                {{ this }}
+            )
+            AND (
+            SELECT
+                LEAST(COALESCE(MAX(block_id), 105368)+9000000,153013616)
+            FROM
+                {{ this }}
+        )
```
desmond-hui, line 49: This batch load logic wont work because `{{ this }}` model does not have a `block_id` column

models/silver/silver__daily_signers.sql
```diff
@@ -0,0 +1,179 @@
...
+                {{ this }}
+        ) 
+    {% elif is_incremental() %}
+        WHERE b_date IN (
+            SELECT
+                block_timestamp_date
+            FROM
+                dates_changed
+        )
+    {% endif %}
```
desmond-hui, line 57: Since batch load logic is needed, use an {% else %} block to give it some small-ish starting range to use when a `--full-refresh` command is given

models/silver/silver__signers.sql
```diff
@@ -0,0 +1,141 @@
+{{ config(
+  materialized = 'incremental',
+  unique_key = "signer",
+  incremental_strategy = 'delete+insert',
+  cluster_by = 'signer'
+) }}
+
+WITH
```
desmond-hui, line 8: I don't think this model can have incremental logic because it is doing an aggregation across the full history of the data.

desmond-hui (changes requested):

models/silver/silver__daily_signers.sql
```diff
@@ -0,0 +1,185 @@
+{{ config(
+  materialized = 'incremental',
+  unique_key = "CONCAT_WS('-', signer, b_date)",
+  incremental_strategy = 'delete+insert',
+  cluster_by = 'signer'
+) }}
+
+WITH 
+{% if is_incremental() %}
```
desmond-hui, line 9: Add an `AND` condition so that this does not run when we are doing a batch load

`env_var(
        'DBT_IS_BATCH_LOAD',
        "false"
    ) == "false"`

desmond-hui (changes requested):

models/silver/silver__daily_signers.sql
```diff
@@ -0,0 +1,188 @@
...
+            SELECT
+                LEAST(COALESCE(MAX(_inserted_timestamp :: date), '2022-08-12')+1, CURRENT_DATE)
+            FROM
+                {{ this }}
+            )
+            AND (
+            SELECT
+                LEAST(COALESCE(MAX(_inserted_timestamp :: date), '2022-08-12')+10, CURRENT_DATE)
+            FROM
+                {{ this }}
```
desmond-hui, line 51: To add to a date you need to use `DATEADD()` function.

Current error when trying to do the batch load

`Can not convert parameter '(SELECT LEAST(DATE_ADDDAYSTODATE(1, IFNULL(MAX(CAST(DAILY_SIGNERS._INSERTED_TIMESTAMP AS DATE)), CAST('2022-08-12' AS DATE))), CURRENT_DATE()) AS "LEAST(COALESCE(MAX(_INSERTED_TIMESTAMP :: DATE), '2022-08-12'::DATE)+1, CURRENT_DATE)" FROM SOLANA_DEV.SILVER.DAILY_SIGNERS AS DAILY_SIGNERS)' of type [DATE] into expected type [NUMBER(38,0)]`

desmond-hui (changes requested):

desmond-hui (changes requested):

models/silver/silver__daily_signers.sql
desmond-hui: `CURRENT_DATE` should be some arbitrary batch load date where the batch load would "stop" and the model would transition to a regular incremental load

models/silver/silver__daily_signers.sql
```diff
@@ -0,0 +1,188 @@
...
+        index,
+        _inserted_timestamp
+    FROM 
+        {{ ref('silver__events2') }} e
+    {% if is_incremental() and env_var(
+        'DBT_IS_BATCH_LOAD',
+        "false"
+    ) == "true" %}
+    WHERE
+        e.block_id BETWEEN (
```
desmond-hui, line 78: This should be `_inserted_timestamp`

desmond-hui (changes requested):

models/silver/silver__daily_signers.sql
```diff
@@ -0,0 +1,254 @@
...
+            LEAST(
+                DATEADD(
+                    'day',
+                    1,
+                    COALESCE(MAX(_inserted_timestamp :: DATE), '2022-08-12')
+                ),
+                CURRENT_DATE
+            )
+        FROM
+            {{ this }}
```
desmond-hui, line 80: I am not quite sure the current flow works either.  If you do a batch load, we probably still want to know which block_timestamps included in that subset so that you can re-aggregate for each one vs. just aggregating that data for that block timestamp date in the subset

desmond-hui (approved):

PR #142, opened 2022-10-22
Fixing bug with Hadeswap NFT Sales

jkallogjeri:


jhuhnke (changes requested):
Not null test on the column "mint" is failing. Here is an example transaction ID of one of the sales that has a null mint value. 

https://solscan.io/tx/3bYDEahhshT7SffhkLwufG5daQkgmMBeNXqC1q4HHdw99SG2Tm6ndBHRihq1qQisUy15i742jHeBif2dfjcca3qi

jhuhnke (approved):

PR #157, opened 2022-11-15
An 2296/intermediate swaps tables 2

tarikceric:
Generalized model for intermediate swaps occurring on dexes (except for Jupiter v4)
- Transfers2 created to include failed transactions


models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+                AND instruction :accounts [1] :: STRING = '2EXiumdi14E9b8Fy62QcA5Uh6WdHS2b38wtSxp72Mibj'
+            )
+            OR program_id IN (
+                '93BgeoLHo5AdNbpqy9bD12dtfxtA5M2fh3rj72bE35Y3',
+                'routeUGWgWzqBWFcrCfv8tritsqukccJPu3q5GPP3xS'
+            )
+        )
+        AND ARRAY_SIZE(
+            e.instruction :accounts
+        ) > 6
```
desmond-hui, line 55: Does this apply generically?  This condition was originally put in for jupiter only

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+        AND e.block_id > 111442741 -- token balances owner field not guaranteed to be populated before this slot
+        AND inner_instruction_program_ids [0] <> 'DecZY86MU5Gj7kppfUCEmd4LbXXuyZH1yHaP2NTqdiZB' --associated with wrapping of tokens
```
desmond-hui, line 57: Does this apply generically? This condition was originally put in for jupiter only

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+        d.program_id IN (
+            '675kPX9MHTjS2zt1qfr1NYHuzeLXfQM9H24wFSUt1Mp8',
+            '5quBtoiQqxF9Jv6KYKctB59NT3gtJD2Y65kdnB1Uev3h',
+            '93BgeoLHo5AdNbpqy9bD12dtfxtA5M2fh3rj72bE35Y3',
+            'routeUGWgWzqBWFcrCfv8tritsqukccJPu3q5GPP3xS'
+        )
+
+    and associated_account is not null
+    
+),
```
desmond-hui, line 159: If we need to have a program specific CTE, then I think it might be a fit to be its own model

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+    UNION
+    SELECT
+        tx_id,
+        instruction :parsed :info :account :: STRING AS associated_account,
+        COALESCE(
+            instruction :parsed :info :source :: STRING,
+            instruction :parsed :info :owner :: STRING
+        ) AS owner
+    FROM
+        solana_dev.silver.events
```
desmond-hui, line 192: This needs to be jinjafied

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+        tx_id,
+        instruction :parsed :info :delegate :: STRING AS delegate,
+        instruction :parsed :info :owner :: STRING AS delegate_owner
+    FROM
+        solana_dev.silver.events
+    WHERE
+        (
+            program_id = 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA'
+            AND event_type = 'approve'
+        )
```
desmond-hui, line 227: Shouldn't this be part of the account mappings CTE?  mapping delegates to the original owner is just another type of account mapping?

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+    AND block_timestamp :: DATE >= '2021-12-14'
+{% endif %}
+),
+delegate_mappings AS(
+    SELECT
+        tx_id,
+        instruction :parsed :info :delegate :: STRING AS delegate,
+        instruction :parsed :info :owner :: STRING AS delegate_owner
+    FROM
+        solana_dev.silver.events
```
desmond-hui, line 222: Also needs to be jinjafied

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+-- AND block_timestamp :: DATE = '2022-07-27'
+AND _inserted_timestamp >= (
+    SELECT
+        MAX(_inserted_timestamp)
+    FROM
+        {{ this }}
+)
+{% else %}
+    AND block_timestamp :: DATE >= '2021-12-14'
+{% endif %}
```
desmond-hui, line 239: Lets consider creating a base events cte that encompasses all the programids we are interested in so that we minimize the number of incremental blocks

desmond-hui (changes requested):

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: jinjafied

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: jinjafied

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: No, this seems to only apply to Jupiter v3/v4 -- moved to only filter out Jupiter swaps

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: Should this logic just be added to the account_mapping union instead of being its own CTE?

desmond-hui (changes requested):

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: It was made separate because it was only used to get the 'min_idx_of_swapper' to identify the first action of the swap (useful in the instance it is out of order) and I don't believe the delegates mapping is necessary in the full account_mapping

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+        solana_dev.silver.events
+    WHERE
+        (
+            program_id = 'ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL'
+            AND event_type = 'create'
+        )
+        OR (
+            program_id = 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA'
+            AND event_type = 'closeAccount'
+        )
```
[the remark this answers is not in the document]
tarikceric: paranthesis added

models/silver/swaps/silver__swaps_intermediate_dex.sql
desmond-hui: Yes, I think that is a good approach

models/silver/swaps/silver__swaps_intermediate_dex.sql
```diff
@@ -0,0 +1,430 @@
...
+        destination,
+        authority,
+        source,
+        min_index_swapper,
+        rn,
+        inner_rn,
+        mint as to_mint, 
+        amount as to_amt
+    from swaps as s1
+    WHERE min_index_swapper is null
```
desmond-hui, line 402: I think these two union statements can be combined using an `OR` since the output columns use the same criteria?  What I mean by that is...both selects hardcode a `NULL` for amount and mint and use s1.amount and s1.mint as the `to_mint` and `to_amt`

desmond-hui (changes requested):

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: base_events CTE has been created and has replaced {{ ref('silver__events') }}. Unnecessary incremental logic has been removed.

models/silver/swaps/silver__swaps_intermediate_dex.sql
tarikceric: The unions have been combined using an `OR` clause

models/silver/swaps/silver__swaps_intermediate_generic.sql
```diff
@@ -0,0 +1,372 @@
+{{ config(
+    materialized = 'incremental',
+    unique_key = "CONCAT_WS('-', block_id, tx_id)",
+    incremental_strategy = 'delete+insert',
+    cluster_by = ['block_timestamp::DATE'],
```
desmond-hui, line 5: Please change this to use the `merge` strategy and unique key should be list `['block_id', 'tx_id']` (see transfers2 for example)

Also please change add `_inserted_timestamp::date` to cluster key

desmond-hui (changes requested):

models/silver/swaps/silver__swaps_intermediate_generic.sql
desmond-hui: Actually, it also needs `swap_index` in the unique key

tarikceric: config has been updated

models/silver/swaps/silver__swaps_intermediate_generic.sql
```diff
@@ -0,0 +1,373 @@
...
+        AND t.block_timestamp :: DATE = e.block_timestamp :: DATE
+    WHERE
+        (
+            program_id IN (
+                -- Orca
+                'MEV1HDn99aybER3U3oa9MySSXqoEZNDEQ4miAimTjaW',
+                '9W959DqEETiGZocYWCQPaJ6sBmUzgfxXfqGeTEdp3aQP',
+                'DjVE6JNiYqPL2QXyCUUh8rNjHrbz9hXHNYt99MQ59qw1',
+                -- saber
+                'Crt7UoUR6QgrFrN7j8rmSQpUTNWNSitSwWvsWGf1qZ5t'
```
desmond-hui, line 70: I think we should revisit these conditions.  Because these are liquidity programs I am pretty sure we are including non-swap transactions in them.  Example

`3PdQf87gUGqRKrTXWzqpuduUVhe2cCfSFbmDbbJWQg4MwmsQu43esBRz3HQPsGZ98a8ZjtcNsxNDsrxhV1LBWNJn`
`5hgT6wotqHT6c49iwJXexnvnCToteo5npbRStjFveMerx4RtR75azqXWKoaFyEWbXmLcEXR68EWwFgj5BzxjjXuy`
`4BhhKEKu1z7qXGPaoX1114oQ5Beqa2VWfgBvoW2dzcU6CmpAGPfztJZim2ZeF6MQsjdAAaiyr4ToChr3fa88mJpA`

models/silver/swaps/silver__swaps_intermediate_generic.sql
```diff
@@ -0,0 +1,373 @@
...
+        t
+        ON t.tx_id = e.tx_id
+        AND t.block_timestamp :: DATE = e.block_timestamp :: DATE
+    WHERE
+        (
+            program_id IN (
+                -- Orca
+                'MEV1HDn99aybER3U3oa9MySSXqoEZNDEQ4miAimTjaW',
+                '9W959DqEETiGZocYWCQPaJ6sBmUzgfxXfqGeTEdp3aQP',
+                'DjVE6JNiYqPL2QXyCUUh8rNjHrbz9hXHNYt99MQ59qw1',
```
desmond-hui, line 68: These orca v1 swaps don't look accurate

`5aSxXzFNBtkeSmuGV3Jq3UAf9B3LEXEZSvB5u2SVoQj3XDmYA7jrMjMK39Gw8eKDjtkYh2CaGtnzB71jn3FMEdT9`
`3hPfLKsSs5dG9GWq3286PGJjTgRrzzPwmRigVcnwVJ3BjVygRHf3TNSf6T1xw6FzaKAaR4TykfuFSuv3PcdWbdMF`

models/silver/swaps/silver__swaps_intermediate_generic.sql
```diff
@@ -0,0 +1,373 @@
...
+        base_events e
+        INNER JOIN {{ ref('silver__transactions') }}
+        t
+        ON t.tx_id = e.tx_id
+        AND t.block_timestamp :: DATE = e.block_timestamp :: DATE
+    WHERE
+        (
+            program_id IN (
+                -- Orca
+                'MEV1HDn99aybER3U3oa9MySSXqoEZNDEQ4miAimTjaW',
```
desmond-hui, line 66: This doesn't seem like it is ORCA exclusive.  May be a custom program that utilizes other swap programs kinda like Jupiter does for public.  Probably want to exclude as we are only dealing w/ "direct" swaps in these models

models/silver/swaps/silver__swaps_intermediate_generic.sql
desmond-hui, line 70: Should add `SSwpkEEcbUqx4vtoEByFjSkhKdCT862DNVb52nZg1UZ` for saber

desmond-hui (changes requested):

models/silver/swaps/silver__swaps_intermediate_generic.sql
tarikceric: Program ID has been removed

models/silver/swaps/silver__swaps_intermediate_generic.sql
tarikceric: These are incorrect due to there being multiple signers and the official swapper is the 2nd one listed -- This is something that was not accounted for and there doesn't seem to be anything unique that could filter these so I will need to explore more possible solutions

PR #160, opened 2022-11-22
feat: Added price feeds for Star Atlas tokens POLIS and ATLAS

dahifi:
Redo for https://github.com/FlipsideCrypto/sql_models/pull/637

So I saw the July 20 cutoff for legacy data, which I suppose means that we won't have data for these tokens beyond that date?


desmond-hui:
@dahifi - We should still have data for those tokens beyond that date, just uses a different data source location.  We will take a look at this and get back to you if any questions or issues.  Otherwise look for this to be merged next week, thanks!

desmond-hui (changes requested):
For `models/silver/prices/silver__token_prices_coin_market_cap_hourly.sql` The new ids are missing from the `base_prices` CTE

dahifi:
Looks like these tokens were added as part of an update Jessica did earlier today. Awaiting update in production data ... later?

PR #166, opened 2022-11-30
prices fix

jhuhnke:


models/silver/prices/silver__token_prices_coin_market_cap_hourly.sql
```diff
@@ -62,13 +62,87 @@ asset_metadata AS (
             7129,
             4195,
             11181, 
-            15610, 
-            16352, 
-            18069, 
-            2396, 
-            3717, 
-            9443, 
-            11212
```
desmond-hui, line 71: Looks like star-atlas got removed?  Not sure what else is was deleted that wasn't meant to be

desmond-hui (changes requested):

models/silver/prices/silver__token_prices_coin_market_cap_hourly.sql
jhuhnke: Ok, added the 5 escapee shitcoins back in

desmond-hui (approved):

PR #176, opened 2022-12-06
AN-2517 cross chain and solscan address to symbol to price

eric-laurello:
1. UDFs for solscan
2. Token Metadata
3. EZ Prices


models/core/core__ez_token_prices.sql
```diff
@@ -0,0 +1,42 @@
...
+    CROSS JOIN base b
+    LEFT JOIN {{ ref('silver__token_prices_coin_gecko_hourly') }}
+    cg
+    ON A.coin_gecko_id = cg.id
+    AND b.recorded_hour = cg.recorded_hour
+    LEFT JOIN {{ ref('silver__token_prices_coin_market_cap_hourly') }}
+    cmc
+    ON A.coin_gecko_id = cmc.id
+    AND b.recorded_hour = cmc.recorded_hour
+WHERE
```
desmond-hui, line 38: Output doesnt look to be showing CMC prices properly.  Example

`select *
from solana_dev.core.ez_token_prices
where token_address = '4k3Dyjzvzp8eMZWUXbBCjEvwSkkk59S5iCNLY3QrkX6R'
order by recorded_hour desc;`

This is Raydium token and is in the silver prices and the token metadata mapping is there

`select *
from solana_dev.silver.token_prices_coin_market_cap_hourly
where id = 8526
order by recorded_hour desc;`

`select *
from solana_dev.silver.token_metadata
where token_address = '4k3Dyjzvzp8eMZWUXbBCjEvwSkkk59S5iCNLY3QrkX6R';`

desmond-hui (changes requested):

models/bronze_api/bronze_api__solscan_token.sql
```diff
@@ -0,0 +1,78 @@
...
+    DATA :data :symbol :: STRING AS symbol,
+    DATA :data :decimals :: INT AS decimals,
+    DATA :data :tag AS tags,
+    DATA :data :icon :: STRING AS icon,
+    DATA :data :twitter :: STRING AS twitter,
+    DATA :data :website :: STRING AS website,
+    _inserted_timestamp,
+    _inserted_timestamp_from_swap
+FROM
+    base
```
desmond-hui, line 78: Shouldn't this be left as raw data and then we transform it in the silver layer?  That would follow the pattern w/ the other macro + our current calls to streamline

models/bronze_api/bronze_api__solscan_token.sql
```diff
@@ -0,0 +1,78 @@
...
+{% endif %}
+GROUP BY
+    A.swap_from_mint
+LIMIT
+    1
+)
+)
+SELECT
+    mint,
+    DATA :data :"address" :: STRING AS token_address,
```
desmond-hui, line 66: These two will always be the same value

models/silver/swaps/silver__swaps.sql
```diff
@@ -0,0 +1,63 @@
+{{ config(
```
desmond-hui, line 1: Please change name of this model.  We are using this in current development as part of the swap models refactoring

models/bronze_api/bronze_api__solscan_token.sql
```diff
@@ -0,0 +1,78 @@
+{{ config(
```
desmond-hui, line 1: This took > 300s to run to get metadata for 1 token? That seems too long, all the incrementals combined only take 15-20mins on a medium.  I am reluctant to add 300s of overhead for 1 token

desmond-hui (changes requested):

models/core/core__dim_tokens.sql
```diff
@@ -0,0 +1,19 @@
+{{ config(
```
desmond-hui, line 1: One of these token addresses is a URL for some reason

`select *
from solana_dev.core.dim_tokens
where symbol = 'COFFEE';`

desmond-hui (changes requested):

models/bronze_api/bronze_api__solscan_token.sql
eric-laurello: some of the calls do not return a value for address so I wanted to keep both for the sake of knowing that.

models/silver/swaps/silver__swaps.sql
eric-laurello: will that new version include the same data? should I merge that change into my branch?

desmond-hui: Same columns but different data.  Probably not too relevant though because I dont see a significant change in terms of the mints that matter

desmond-hui (approved):

PR #210, opened 2023-01-12
An 2661/saber lp actions

tarikceric:
Model for actions (withdraw/deposits) performed on Saber dex liquidity pools.


desmond-hui (changes requested):
Can you add an exclude for this model from the regular incremental workflow?  Then after this is merged + full refreshed in prod add a PR to include it back.

desmond-hui (changes requested):
How did you calculate TVL for a pool like '7oAd7xG4m3oC2qeWB1szghTebAZyyGPAFDJ4wwwbRSNi'?

desmond-hui (approved):

PR #216, opened 2023-01-20
capture additional metaplex mints

tarikceric:
Add logic to capture missing Metaplex mints, which is causing a number of issues in downstream NFT models. These additional mints are captured from within `instructions` and `inner_instructions`, and these mints are relatively new with the first one appearing on 2022-09-28.


desmond-hui (changes requested):
Please add test-cases to `models/silver/nfts/silver__nft_mints.yml`

desmond-hui (approved):

PR #220, opened 2023-01-26
adjust incremental date range

tarikceric:
Temporarily adjust date range for incremental run in GH actions


desmond-hui (changes requested):
This is not necessary

PR #230, opened 2023-02-01
An 2743/liquidity pools

tarikceric:
Dim table containing information for liquidity pools on Orca. `WhirLbMiicVdio4qvUfM5KAg6Ct8VwpYzGff3uctyCc` pools mint unique NFTs when a deposit is made to represent the LP amount, so the pool_token column is null for these.

### DIFF
--- a/models/silver/liquidity_pool/silver__dim_orca_liquidity_pools.sql
+++ b/models/silver/liquidity_pool/silver__dim_orca_liquidity_pools.sql
@@ -52,8 +52,8 @@ orca_pools AS (
         l.liquidity_pool,
         l.owner,
         e.instruction :accounts [1] :: STRING AS mint_authority,
-        e.instruction :accounts [2] :: STRING AS token_a,
-        e.instruction :accounts [3] :: STRING AS token_b,
+        e.instruction :accounts [2] :: STRING AS token_a_account,
+        e.instruction :accounts [3] :: STRING AS token_b_account,
         e.instruction :accounts [4] :: STRING AS pool_token,
         l._inserted_timestamp
     FROM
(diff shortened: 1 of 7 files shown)


models/silver/liquidity_pool/silver__dim_orca_liquidity_pools.sql
```diff
@@ -0,0 +1,98 @@
+{{ config(
```
desmond-hui, line 1: So I think if we can leave out the `dim_` at the silver level and maybe just call this `silver__liquidity_pool_creations_orca` <-- also changing the word order to match our standard (see how we named the intermediate swap pools)

models/silver/liquidity_pool/silver__dim_orca_liquidity_pools.sql
```diff
@@ -0,0 +1,98 @@
...
+    materialized = 'incremental',
+    unique_key = ["block_id","tx_id"],
+    incremental_predicates = ['DBT_INTERNAL_DEST.block_timestamp::date >= LEAST(current_date-7,(select min(block_timestamp)::date from ' ~ generate_tmp_view_name(this) ~ '))'],
+    cluster_by = ['block_timestamp::DATE','_inserted_timestamp::DATE'],
+    post_hook = "ALTER TABLE {{ this }} ADD SEARCH OPTIMIZATION"
```
desmond-hui, line 6: Remove this post_hook, no need for search optimization on this table

models/silver/liquidity_pool/silver__dim_orca_liquidity_pools.yml
```diff
@@ -0,0 +1,51 @@
...
+        tests: 
+          - not_null 
+      - name: TOKEN_B_Account
+        description: "Token address for second token used in the liquidity pool"
+        tests: 
+          - not_null 
+      - name: POOL_TOKEN
+        description: "Address for pool token associated with the liquidity pool"
+        tests: 
+          - not_null
```
desmond-hui, line 47: I believe you said these can be non-null for whirlpools, so we should condition this test

desmond-hui (changes requested):

desmond-hui (approved):

PR #233, opened 2023-02-03
An 2745/burn and mint actions

tarikceric:
- Add `signer`/`mint_authority` columns to `mint_actions` table
- Create a `burn_actions` table


models/silver/silver__burn_actions.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: MINT
+        description: "{{ doc('mint') }}"
+        tests: 
+          - not_null
+      - name: BURN_AMOUNT
+        description: Number of tokens burned
+        tests: 
+          - not_null
+      - name: BURN_AUTHORITY
+        description: Account address authorizing burn
```
desmond-hui, line 67: this should be non-null always?

models/silver/silver__mint_actions.sql
```diff
@@ -56,6 +56,11 @@ SELECT
         instruction :parsed :info :amount :: INTEGER,
         instruction :parsed :info :tokenAmount: amount :: INTEGER
     ) AS mint_amount,
+    COALESCE(
+        instruction :parsed :info :mintAuthority :: string,
+        instruction :parsed :info :multisigAuthority :: string
+    ) AS mint_authority,
+    instruction :parsed :info :signers[0] :: string AS signer,
```
desmond-hui, line 63: Why are we just taking the first one instead of outputting the entire array?

models/silver/silver__burn_actions.sql
```diff
@@ -0,0 +1,99 @@
...
+    instruction :parsed :info :mint :: STRING AS mint,
+    COALESCE(
+        instruction :parsed :info :amount :: INTEGER,
+        instruction :parsed :info :tokenAmount: amount :: INTEGER
+    ) AS burn_amount,
+    COALESCE(
+        instruction :parsed :info :authority :: string,
+        instruction :parsed :info :multisigAuthority :: string
+    ) AS burn_authority,
+    instruction :parsed :info :signers[0] :: string AS signer,
```
desmond-hui, line 63: Why are we just taking the first one instead of outputting the entire array?

desmond-hui (changes requested):

models/silver/silver__burn_actions.yml
tarikceric: That should be true -- updated to include not_null test

models/silver/silver__mint_actions.sql
tarikceric: Thats an assumption on my part that the first item is the only relevant one, but it's a much better idea to grab the whole array -- updated

desmond-hui (approved):

PR #237, opened 2023-02-07
An 2149/account owners model

desmond-hui:
- Add model to capture token account ownership events
- Add model to capture stake account ownership (staker or withdrawer) events
- Currently does not handle vote account ownerships
- Add aggregated model for token account ownership
- Add backfill helper macro, this will handle load times for uneven raw data distribution and should be used for larger tables


models/silver/accounts/silver__stake_account_ownership_events.sql
```diff
@@ -0,0 +1,164 @@
...
+        index,
+        inner_index,
+        event_type,
+        instruction:parsed:info:stakeAccount::string as account_address,
+        lower(instruction:parsed:info:authorityType::string) as authority_type,
+        instruction:parsed:info:newAuthority::string as authority,
+        _inserted_timestamp
+    from ownership_change_events 
+    where event_type in ('authorize','authorizeChecked')
+    and instruction:parsed:info:authorityType::string = 'Withdrawer' /* probably handle stake accounts differently and include both stake and withdraw authorities */
```
tarikceric, line 64: The authority_type values for withdrawer/staker are inconsistent with some having the first letter capitalized

models/silver/accounts/silver__stake_account_ownership_events.sql
```diff
@@ -0,0 +1,164 @@
...
+        index,
+        inner_index,
+        event_type,
+        instruction:parsed:info:stakeAccount::string as account_address,
+        lower(instruction:parsed:info:authorityType::string) as authority_type,
+        instruction:parsed:info:newAuthority::string as authority,
+        _inserted_timestamp
+    from ownership_change_events 
+    where event_type in ('authorize','authorizeChecked')
+    and instruction:parsed:info:authorityType::string = 'Staker' /* probably handle stake accounts differently and include both stake and withdraw authorities */
```
tarikceric, line 81: for some txs like `5kEzA1C3BxAdKpQv3RtcnJRREvLToXs326aJiHuZH698JdHrRLeSNnVkFH8uCjjk15G3oSwwGsnru1916HuJFxPL` which include both withdrawer + staker events, should these be separate records or is it ok that only `staker` is present

tarikceric: Similarly for a tx like `5o71fKvXPcDvpEWuTUZk5uKEoHvkmRdvVeYKpQVtBJ2FS18iYuewQwn2QWvv7DHzEvqABrtG7BkMJa1QzW79ztkz` - does it matter if its a `withdrawer` instead of `staker`?

models/silver/accounts/silver__token_account_ownership_events.sql
```diff
@@ -0,0 +1,163 @@
...
+    where succeeded 
+    {% if is_incremental() %}
+        {% if execute %}
+        {{ get_batch_load_logic(this,15,'2023-02-05') }}
+        {% endif %}
+    {% else %}
+        and _inserted_timestamp::date between '2022-08-12' and '2022-09-01'
+    {% endif %}
+),
+ownership_change_events as (
```
tarikceric, line 21: There appear to be some txs with duplicates such as `xYs6tZo9aVLWS2PcmrzgcpFrtFgCBpULK8Wu78L8Ymo7Lwe1M1qnaPxUucEfq6zEWyTDEBKT6VrReRDmzGWLkRj`

models/silver/accounts/silver__token_account_ownership_events.yml
```diff
@@ -0,0 +1,50 @@
...
+          - not_null
+      - name: ACCOUNT_ADDRESS
+        description: address of token account
+        tests: 
+          - not_null
+      - name: OWNER
+        description: address of new owner
+        tests: 
+          - not_null
+      - name: MINT
```
tarikceric, line 43: `null` test fails as there are tx's with null mints

models/silver/accounts/silver__token_account_owners.sql
```diff
@@ -0,0 +1,74 @@
...
+    unique_key = ["account_address","owner","start_block_id"],
+    cluster_by = ['_inserted_timestamp::DATE'],
+) }}
+
+/*
+for incrementals also select all null end date accounts and combine
+join to eliminate accounts that are not in the subset
+remove all accounts that have the same owner + start block + end block
+*/
+with last_updated_at as (
```
tarikceric, line 12: There are some situations where different ownership periods have the same end_block_id, ie for `account_address = 'Hq5cBoPxUAT5KFvQYWJXMDWo7tosUYQaHGzs2Bz48su7'` -- does this make sense?

models/silver/accounts/silver__token_account_owners.yml
```diff
@@ -0,0 +1,22 @@
...
+          - not_null
+      - name: OWNER
+        description: address of owner
+        tests:
+          - not_null
+      - name: START_BLOCK_ID
+        description: block where this ownership begins 
+        tests:
+          - not_null
+      - name: END_BLOCK_ID
```
tarikceric, line 17: Some accounts_addresses have multiple records where the end date is null ie `rbfxAVZbCPRoeEf4Bfwy1P43gm6w3jaP3iyn4jF1wHq`

models/silver/accounts/silver__token_account_ownership_events.yml
desmond-hui: This should be a non-issue.  Here is the note from original slack message

```mint not null test will fail because I only added it this morning but didnt want to redo a full refresh on the dev table so only that incremental will have it filled out```

File: models/silver/accounts/silver__stake_account_ownership_events.sql (hunk 2 above)
Comment (desmond-hui) [reply models/silver/accounts/silver__stake_account_ownership_events.sql:81]: Good catch, it matters and both should exist.  I think adding the `authority_type` to qualify will fix

File: models/silver/accounts/silver__stake_account_ownership_events.sql (hunk 1 above)
Comment (desmond-hui) [reply models/silver/accounts/silver__stake_account_ownership_events.sql:64]: Going to add `lower()` to make them all consistent casing

File: models/silver/accounts/silver__token_account_owners.yml (hunk 6 above)
Comment (desmond-hui) [reply models/silver/accounts/silver__token_account_owners.yml:17]: OK, i need to figure out some logic to handle this situation because it's being caused by the data coming out of order

Comment (desmond-hui) [reply models/silver/accounts/silver__token_account_owners.yml:17]: ...and it looks like this account gets created over and over again...

File: models/silver/accounts/silver__token_account_ownership_events.sql (hunk 3 above)
Comment (desmond-hui) [reply models/silver/accounts/silver__token_account_ownership_events.sql:21]: I think this has to do with me loading data before changing `inner_index` to be a non-null value.  Probably created duplicates instead of replacing.

File: models/silver/accounts/silver__token_account_owners.sql (hunk 5 above)
Comment (desmond-hui) [reply models/silver/accounts/silver__token_account_owners.sql:12]: Same w/ this one, looks like unintended effect of data coming out of order

File: .github/workflows/dbt_run_incremental.yml [hunk 7]
```diff
@@ -44,6 +44,6 @@ jobs:
...
           dbt run-operation run_sp_refresh_external_tables_full
-          dbt run -s ./models --exclude models/core models/silver/silver__transactions.sql models/silver/silver__blocks.sql models/silver/silver__votes.sql models/silver/silver___inner_instructions.sql models/silver/silver___instructions.sql models/silver/silver__events.sql models/silver/silver___all_undecoded_instructions_data.sql tag:share models/streamline models/silver/silver__daily_signers.sql models/silver/silver__signers.sql
+          dbt run -s ./models --exclude models/core models/silver/silver__transactions.sql models/silver/silver__blocks.sql models/silver/silver__votes.sql models/silver/silver___inner_instructions.sql models/silver/silver___instructions.sql models/silver/silver__events.sql models/silver/silver___all_undecoded_instructions_data.sql tag:share models/streamline models/silver/silver__daily_signers.sql models/silver/silver__signers.sql models/silver/accounts/silver__token_account_ownership_events.sql models/silver/accounts/silver__token_account_owners.sql models/silver/accounts/silver__stake_account_ownership_events.sql
```
tarikceric, line 47: should the new incremental model be included as well?

models/silver/accounts/silver__token_account_owners.sql
```diff
@@ -0,0 +1,48 @@
...
+select 
+    account_address,
+    owner,
+    start_block_id,
+    lead(start_block_id) ignore nulls over (
+                PARTITION BY account_address
+                ORDER BY bucket
+            ) as end_block_id,
+    _inserted_timestamp
+from regroup
```
tarikceric, line 47: For this acct '`EUfmV6DzpSUi3rGjZmjbZtKsMWHYoZjCV9V5cB4uMi1s`'  it looks like the current owner is `TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA` but within solscan it says its still `aGNoP1WF8Mbs8mwEBVcgM1kkKnX3R4myrE2sZifUmfe` -- where the data currently says the owner only held it temporarily

models/silver/accounts/silver__stake_account_ownership_events.sql
```diff
@@ -0,0 +1,164 @@
+{{ config(
```
tarikceric, line 1: Just curious if there's a plan to include a `stake_account_owners` table as well

.github/workflows/dbt_run_incremental.yml
desmond-hui: Good call, need to do that!

models/silver/accounts/silver__token_account_owners.sql
desmond-hui: Taking a look at root cause, solscan looks right

desmond-hui: Ok...this one is annoying, there are change events happening in diff transactions but they are in the same block so it becomes out of order and thus causing some weird data outputs

desmond-hui: Alright, I think the only way to add some ordering based on the event type to prevent this from happening because we don't have an order for the transactions themselves.  I added this and reloaded

tarikceric (approved):

PR #238, opened 2023-02-07
correct string to 'multisigMintAuthority'

tarikceric:
Correct the key being searched for from 'multisigAuthority' to 'multisigMintAuthority'

### DIFF
--- a/models/silver/silver__burn_actions.sql
+++ b/models/silver/silver__burn_actions.sql
@@ -57,7 +57,7 @@ SELECT
     ) AS burn_amount,
     COALESCE(
         instruction :parsed :info :authority :: string,
-        instruction :parsed :info :multisigMintAuthority :: string
+        instruction :parsed :info :multisigAuthority :: string
     ) AS burn_authority,
     instruction :parsed :info :signers :: string AS signers,
     _inserted_timestamp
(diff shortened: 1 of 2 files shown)


models/silver/silver__burn_actions.sql
```diff
@@ -57,7 +57,7 @@ SELECT
     ) AS burn_amount,
     COALESCE(
         instruction :parsed :info :authority :: string,
-        instruction :parsed :info :multisigAuthority :: string
+        instruction :parsed :info :multisigMintAuthority :: string
```
desmond-hui, line 60: This doesn't seem like the right fix, I am seeing results where `multisigAuthority` is used

`select
    instruction :parsed :info :multisigAuthority :: string as m_sig
from solana_dev.silver.events
where event_type IN (
       'burn',
        'burnChecked'
    )
and m_sig is not null;`

desmond-hui (changes requested):

models/silver/silver__burn_actions.sql
tarikceric: You are spot on per usual - I couldn't find `multisigMintAuthority` for burns and vice versa so just reverted the burn_actions table

desmond-hui (approved):

PR #247, opened 2023-02-13
An 1904/user interactions

jhuhnke:
Tested with my burner wallet: 2L6j3wZXEByg8jycytabZitDh9VVMhKiMYv7EeJh6R2H

Two known edge cases: 

1. The Solana Spaces DRiP airdrop NFTs are supposed to be minted directly into 2L6... and other wallets it is airdropped to. However, a different wallet signs the mint transaction and thus these NFTs are not covered by our silver mints model. They do not appear in transfers as well. Example mint transaction: https://solscan.io/tx/3CB8mcdpKjAKKGiyWaHeLE6BU1hzVRZWnQck6CbpvE65WjUscp71wMHYrC4uT9ciJFNHZnnmTTpLJHu6CRBuSNUF

2. Some Magic Eden V2 sales seem to transfer ownership of the NFT accounts, yet do not appear in the token account ownership events table. This seems to occur only when the token is listed on Magic Eden directly via the Phantom wallet interface. This may be fixed with the bug fix just put in by Desmond for null mints, but my workaround was to also include the Magic Eden V2 silver model in the tokens out CTE. Here's an example of a token I minted and then sold via the Phantom wallet interface: https://solscan.io/token/2r4HsVHCDFbNhysMiHri44fV8YFuVH47PPe23u8bKcPW


models/silver/silver___daily_signers_nft_change.sql
```diff
@@ -0,0 +1,263 @@
...
+    incremental_strategy = 'delete+insert',
+    cluster_by = 'signer'
+) }} 
+WITH dates_changed AS (
+    SELECT
+        DISTINCT block_timestamp :: DATE AS block_timestamp_date
+    FROM
+        {{ ref('silver__transactions') }}
+
+{% if is_incremental() %}
```
tarikceric, line 13: Encountering some issues running an incremental

models/silver/silver___daily_signers_nft_change.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+models:
+  - name: silver__daily_signers_nft_change
```
tarikceric, line 3: looks like we need an extra '_' to 'silver___daily_signer...' to match the correct spelling and be able to run tests

models/silver/silver___daily_signers_nft_change.sql
```diff
@@ -0,0 +1,263 @@
...
+        {% if execute %}
+        {{ get_batch_load_logic(this,30,'2023-02-14') }}
+        {% endif %}
+    {% else %}
+        AND _inserted_timestamp::date between '2023-02-05' and '2023-02-14'
+    {% endif %} 
+
+), 
+ins AS (
+    SELECT
```
tarikceric, line 224: Theres instances where nfts_in or nfts_out are empty arrays [], and other times when they are NULL -- do you think this will have some impact?

models/silver/silver___daily_signers_nft_change.yml
jhuhnke: Good catch - I think I changed this back and forth too much and forgot :sweat_smile:

models/silver/silver___daily_signers_nft_change.sql
jhuhnke: Good catch - the macro for the batch logic incremental isn't set up yet for cases where there are multiple _inserted_timestamps from joining tables.

models/silver/silver___daily_signers_nft_change.sql
jhuhnke: Tested this using two signers that contained both [] and nulls in the nft_outs and found no difference: 
`WHERE
        signer IN ('11aVYZ5rySnGmRghjbnmX6RX1rbN1VhahDojCK395Jy', '11169iRGU9wo4A8MAXVB4ffbkYqJpwaHhYw6Tp88LV7')`
        
Did find two different small bugs while testing though, so ended up being a good exercise

models/silver/silver___daily_signers_nft_change.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+models:
+  - name: silver___daily_signers_nft_change
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - signer
+            - b_date
+    columns:
+      - name: SIGNER
```
tarikceric, line 10: got some NULL signers when running the tests

jhuhnke: Ok - no need to keep null signers in here. I'm going to just remove these.

tarikceric (approved):

PR #254, opened 2023-02-17
An 1904/user interactions

jhuhnke:


models/silver/silver__mint_actions.sql
```diff
@@ -69,7 +48,8 @@ WHERE
         'mintTo',
         'initializeMint',
         'mintToChecked',
-        'initializeMint2'
+        'initializeMint2', 
+        'create'
```
tarikceric, line 52: Curious on the addition of 'create' events -- does this have a decent impact on the current 'mint_actions' data? and do these events always relate to some sort of mint action?

models/silver/silver__mint_actions.sql
```diff
@@ -2,45 +2,23 @@
...
-            MAX(_inserted_timestamp)
-        FROM
-            {{ this }}
-    )
-{% else %}
-WHERE
-    block_id between 31310775 and 32310775
-{% endif %}
+        {{ ref('silver__events') }} 
+    WHERE succeeded
```
tarikceric, line 14: Since we were including failed events before, I'm just wondering if filtering to only include `succeeded` ones will somehow impact the downstream models -- what are your thoughts?

models/core/core__ez_signers.sql
```diff
@@ -11,6 +11,7 @@ SELECT
     num_days_active,
     num_txs,
     total_fees,
-    programs_used
+    programs_used, 
+    nfts_held
```
tarikceric, line 15: since the `nfts_held` has been removed from the `signers` table, should this be revised or am I maybe missing something

models/silver/silver__mint_actions.sql
jhuhnke: adding in a 1=1 clause so the incremental macro can still run and removing this - people might want to see how many times they failed to mint a hyped project

models/silver/silver__mint_actions.sql
jhuhnke: Removed - I think I convinced myself I needed it for an edge case with the solana spaces NFTs, but checking again it looks like I don't. Here's a spaces NFT: https://solscan.io/tx/Vf68hQibVYCPrCqriznuvNpwAf7kqqdrHTU15XaC6FceGzpgubxWMCBNcQXLreo5fcr2NLYRB79Bxz4hYwTC9Y9

This totally has the 'mintTo' event type

models/core/core__ez_signers.sql
jhuhnke: You aren't missing anything here - I was the dummy

PR #311, opened 2023-06-30
block_production model

tarikceric:
- Create ingestion UDF for daily snapshot of block production data
-  table for historical block production data sent from Solana Foundation

### DIFF
--- a/macros/bronze_api/get_block_production.sql
+++ b/macros/bronze_api/get_block_production.sql
@@ -1,6 +1,15 @@
 {% macro get_block_production() %}
+ {% set create_table %}
+CREATE TABLE if NOT EXISTS {{ target.database }}.bronze_api.block_production(
+    json_data VARIANT,
+    epoch varchar,
+    calls string,
+     _inserted_timestamp timestamp_ntz
+  );
+{% endset %}
+  {% do run_query(create_table) %}
   {% set create_block_prod_call_query %}
-  CREATE temporary TABLE block_production_call AS
+  CREATE temporary TABLE bronze_api.block_production_call AS
 SELECT
   ARRAY_AGG(
     { 'id': previous_epoch,
@@ -39,16 +48,17 @@ FROM
   {% do run_query(create_block_prod_call_query) %}
   {% set results_query %}
 INSERT INTO
-  solana_dev.bronze_API.block_production WITH results AS (
+    bronze_api.block_production (json_data,epoch,calls,_inserted_timestamp) WITH results AS (
     SELECT
       ethereum.streamline.udf_json_rpc_call(
         'https://api.mainnet-beta.solana.com',{ 'Content-Type': 'application/json' },
         calls
       ),
-      calls [0] :id AS epoch,
+      calls [0] :id,
+      calls :: string,
       TO_TIMESTAMP_NTZ(CURRENT_TIMESTAMP) AS _inserted_timestamp
     FROM
-      block_production_call
+      bronze_api.block_production_call
   )
 SELECT
   *
--- a/macros/bronze_api/get_block_production.sql
+++ b/macros/bronze_api/get_block_production.sql
@@ -1,0 +1,59 @@
+{% macro get_block_production() %}
+  {% set create_block_prod_call_query %}
+  CREATE temporary TABLE block_production_call AS
+SELECT
+  ARRAY_AGG(
+    { 'id': previous_epoch,
+    'jsonrpc': '2.0',
+    'method': 'getBlockProduction',
+    'params': [ { 'commitment': 'confirmed', 'range': { 'firstSlot': previous_epoch_start, 'lastSlot': previous_epoch_end } }] }
+  ) calls
+FROM
+  (
+    WITH current_epoch_data AS (
+      SELECT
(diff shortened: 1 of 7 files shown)


macros/bronze_api/get_block_production.sql
```diff
@@ -0,0 +1,59 @@
...
+    )
+    SELECT
+      *
+    FROM
+      temp
+  ) {% endset %}
+  {% do run_query(create_block_prod_call_query) %}
+  {% set results_query %}
+INSERT INTO
+  solana_dev.bronze_API.block_production WITH results AS (
```
eric-laurello, line 42: can this be replaced with a ref?

models/silver/validator/silver__historical_block_production.sql
```diff
@@ -0,0 +1,17 @@
...
+
+SELECT
+    a.json_data :metadata :epoch AS epoch,
+    f.value :identityPubkey :: STRING AS node_pubkey,
+    f.value :leaderSlots :: INT AS num_leader_slots,
+    f.value :blocksProduced :: INT AS num_blocks_produced,
+    a.json_data :metadata :start_slot :: INT AS start_slot,
+    a.json_data :metadata :end_slot :: INT AS end_slot
+FROM
+    solana_dev.bronze.block_production a,
```
eric-laurello, line 13: same here with the ref

models/silver/validator/silver__snapshot_block_production.sql
```diff
@@ -0,0 +1,26 @@
...
+
+select
+    epoch,
+    key as node_pubkey,
+    f.value[0] as num_leader_slots,
+    f.value[1] as num_blocks_produced,
+    json_data:data[0]:result:value:range:firstSlot as start_slot,
+    json_data:data[0]:result:value:range:lastSlot as end_slot,
+    _inserted_timestamp
+from solana_dev.bronze_api.block_production a,
```
eric-laurello, line 16: here too

macros/bronze_api/get_block_production.sql
```diff
@@ -0,0 +1,59 @@
+{% macro get_block_production() %}
+  {% set create_block_prod_call_query %}
+  CREATE temporary TABLE block_production_call AS
```
eric-laurello, line 3: probably worthwhile to put a schema here and on line 51

macros/bronze_api/get_block_production.sql
```diff
@@ -0,0 +1,59 @@
...
+  {% do run_query(create_block_prod_call_query) %}
+  {% set results_query %}
+INSERT INTO
+  solana_dev.bronze_API.block_production WITH results AS (
+    SELECT
+      ethereum.streamline.udf_json_rpc_call(
+        'https://api.mainnet-beta.solana.com',{ 'Content-Type': 'application/json' },
+        calls
+      ),
+      calls [0] :id AS epoch,
```
eric-laurello, line 48: might want to add the full call to the results to log it for future troubleshooting.

macros/bronze_api/get_block_production.sql
```diff
@@ -0,0 +1,59 @@
+{% macro get_block_production() %}
```
eric-laurello, line 1: Probably want to add the create table if doesn't exist command as step one to make sure the table gets created in prod first

macros/bronze_api/get_block_production.sql
tarikceric: great point, will add

macros/bronze_api/get_block_production.sql
tarikceric: another great point, will add that in

macros/bronze_api/get_block_production.sql
tarikceric: looked through a few other similar macros on other chains and noticed they all used the set-up of 'bronze_api.table_name' so went with that, bc I'm not sure the  {{ref (xxx) }} would work since the bronze_api is a source?

macros/bronze_api/get_block_production.sql
tarikceric: added schema

models/silver/validator/silver__historical_block_production.sql
```diff
@@ -0,0 +1,20 @@
+{{ config (
+    materialized = 'view'
+) }}
+
```
eric-laurello, line 4: only one small thing to add. If the macro is run twice during the same epoch would we see dupes in this view? Can you either add a qualify or dedupe into a table?

eric-laurello: rather, not the macro, but whatever is populating this

tarikceric: added in a qualify for the two silver models

this specific model showcases the historical data received from the Foundation which would be a one time pull and shouldn't have dupes, but I think thats a great point to add in the qualify

Also added it in for the snapshot model, since you are correct that running it twice in an epoch would produce duplicates. I believe the incremental_strategy would just overwrite any existing dupes, but I added it in just in case.

eric-laurello (approved):

PR #313, opened 2023-07-06
fix mev2 sell txs

tarikceric:
Fix logic for AMM sales performed via Magic Eden
- separate sell and buy txs

### DIFF
--- a/models/silver/nfts/silver__nft_sales_amm_sell.sql
+++ b/models/silver/nfts/silver__nft_sales_amm_sell.sql
@@ -117,7 +117,7 @@ mev2_sales AS(
             instruction :accounts [12] :: STRING = '11111111111111111111111111111111'
             OR instruction :accounts [16] :: STRING = 'ocp4vWUzA2z2XMYJ3QhM9vWdyoyoQwAFJhRdVTbvo9E'
         )
-    UNION
+    UNION ALL
         -- mev2_sells
     SELECT
         A.*,
--- a/models/silver/nfts/silver__nft_sales_amm_sell.sql
+++ b/models/silver/nfts/silver__nft_sales_amm_sell.sql
@@ -98,11 +98,32 @@ coral_cube_sales AS(
         ) >= 16
 ),
 mev2_sales AS(
+    -- mev2_buys
     SELECT
         A.*,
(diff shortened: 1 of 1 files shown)


models/silver/nfts/silver__nft_sales_amm_sell.sql
```diff
@@ -98,19 +98,42 @@ coral_cube_sales AS(
...
+    WHERE
+        signers [1] = 'NTYeYJ1wr4bpM5xo6zx5En44SvJFAd35zTxxNoERYqd'
+        AND ARRAY_SIZE(
+            instruction :accounts
+        ) > 16
+        AND (
+            instruction :accounts [12] :: STRING = '11111111111111111111111111111111'
+            OR instruction :accounts [16] :: STRING = 'ocp4vWUzA2z2XMYJ3QhM9vWdyoyoQwAFJhRdVTbvo9E'
+        )
+    UNION
```
eric-laurello, line 120: Can you change this to a union all?

models/silver/nfts/silver__nft_sales_amm_sell.sql
```diff
@@ -159,7 +184,14 @@ mev2_nft_sale_amount AS (
         LEFT OUTER JOIN base_transfers b
         ON A.tx_id = b.tx_id
     WHERE
-        A.instruction :accounts [5] = b.tx_from
+        (
+            A.nft_sale_type = 'sell'
+            AND A.instruction :accounts [0] = b.tx_from
+        )
+        OR (
```
eric-laurello, line 191: how does this model run compared the old version? More curious than anything because the OR statement can really slow things down in some cases

models/silver/nfts/silver__nft_sales_amm_sell.sql
tarikceric: Yes it can - revised

models/silver/nfts/silver__nft_sales_amm_sell.sql
tarikceric: it seems to have a negligible effect on run time, since the data being queried isn't relatively large in this case

eric-laurello (approved):

PR #316, opened 2023-07-17
Add solana parser jobs and models

xiuy001:


WHYTEWYLL (approved):

models/sources.yml
```diff
@@ -36,7 +36,7 @@ sources:
       - name: prod_nft_metadata_uploads_1828572827
   - name: bronze_streamline
     database: streamline
-    schema: solana
```
tarikceric, line 39: think we can just revert this back to not use dev - had to test using `solana` to get past stage issue  in dev so should be good

models/bronze/streamline/bronze__streamline_FR_program_parser.sql
```diff
@@ -0,0 +1,11 @@
+{{ config (
+    materialized = 'view'
+) }}
+
+{% set model = this.identifier.split("_") [-2] + '_' + this.identifier.split("_") [-1] %}
+{{ streamline_external_table_query(
+    model,
+    partition_function = "CAST(SPLIT_PART(SPLIT_PART(file_name, '/', 4), '_', 1) AS INTEGER )",
+    partition_name = "_partition_by_block_id",
+    unique_key = "block_id"
```
tarikceric, line 10: other_cols passed here per Eric's pr

PR #318, opened 2023-07-25
An 3519/core validator tables

tarikceric:
- Create core tables from silver validator data
- adjust cols to array/object as needed

### DIFF
--- a/models/gold/core__fact_block_production.sql
+++ b/models/gold/core__fact_block_production.sql
@@ -1,0 +1,22 @@
+{{ config(
+      materialized='view'
+    ) 
+}}
+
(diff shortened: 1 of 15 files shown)


charlieflipside:
Looks good. Confirming, start slot / end slot is the same for the entire epoch right? Or is it at the validator level? (i.e., epoch 450, all vote_pubkey have exact same start slot & end slot?)

charlieflipside (approved):
Approved

tarikceric:
> 

Yes thats correct, every validator in an epoch will have the same `start_slot` + `end_slot`

PR #322, opened 2023-08-02
seperate job for block_production

tarikceric:
Create an individual job for running block_production workflow every 4 hrs


forgxyz (approved):
LGTM

PR #346, opened 2023-09-11
An 3915/reduce model run times

desmond-hui:
- Add incremental logic for `ez_staking_lp_actions`
  - Main change was to eliminate the need to do a full `FIRST_VALUE/LAST_VALUE` on the entire dataset
  - Reason for change was to improve model update performance.  This should bring time down from >5mins to <20secs on a M based on dev tests
  - A separate PR will be made to add some data validation tests.  I want to get this change in so we can get runtime down further in order to enable other changes.


models/gold/gov/gov__ez_staking_lp_actions.sql
```diff
@@ -140,7 +162,8 @@ tx_base AS (
             WHEN event_type = 'initialize' THEN instruction :parsed :info :authorized :withdrawer :: STRING
             WHEN event_type = 'authorize'
             AND instruction :parsed :info :authorityType = 'Withdrawer' THEN instruction :parsed :info :newAuthority :: STRING
-            ELSE instruction :parsed :info :stakeAuthority :: STRING
+            WHEN event_type = 'split_destination' THEN stake_authority
```
eric-laurello, line 165: Is this a logic change too?

models/gold/gov/gov__ez_staking_lp_actions.sql
```diff
@@ -311,7 +302,7 @@ temp2 AS (
         withdraw_amount,
         withdraw_destination,
         CASE
-            WHEN vote_account IS NULL THEN FIRST_VALUE(vote_account) ignore nulls over (
+            WHEN vote_account IS NULL THEN LAST_VALUE(vote_account) ignore nulls over (
```
eric-laurello, line 305: same here. This is an intended logic change too?

eric-laurello: Looks like this change is nulling out the vote account and validator fields:
`select   * from solana_dev.gov.ez_staking_lp_actions
WHERE block_timestamp::date >= current_date - 1 and tx_id = '3QqjbHo7M9fvhZvuM7ir1EXbzA93LXWdKKzfwARCzaFvQ52fpbm3NyWJ1QHnsWgEnQkio1iAf43sn1EFTDQh8dXx' and index = 0.2
UNION ALL
select   * from solana.gov.ez_staking_lp_actions
WHERE block_timestamp::date >= current_date - 1 and tx_id = '3QqjbHo7M9fvhZvuM7ir1EXbzA93LXWdKKzfwARCzaFvQ52fpbm3NyWJ1QHnsWgEnQkio1iAf43sn1EFTDQh8dXx' and index = 0.2`

desmond-hui: @eric-laurello - this is intended logic change, the current table is attributing vote accounts to previous txns before they are actually delegated.  You can see the sequence for a given stake account with query below.  I checked a good amount of these and new table logic was always correct from my samples.

`select block_id,tx_id,event_type, stake_account, stake_authority, withdraw_authority, vote_account
from solana_dev.gov.ez_staking_lp_actions
where stake_account in ('CLkomm8TfxBQ6xYebwHvbHc9AFvdHNdWsoBXxf7zuguD')
order by block_id desc, index desc;`

models/gold/gov/gov__ez_staking_lp_actions.sql
desmond-hui: @eric-laurello - intended logic change, same as other explanation.

tarikceric (approved):

PR #367, opened 2023-10-19
add full_index col for unique test

tarikceric:
- The previous unique_key was faulty due to concatenation of` NULL` `inner_index` values, which results in a `NULL` unique key and impacts incremental runs
- created a `full_index` column by concatenating the `index` + `inner_index` as strings and used that in the unique key


desmond-hui (changes requested):
I think this implementation is fine in a vaccuum...but there are already models handling this using different logic, so I think we should stick to the existing implementation instead of having multiple solutions for the same problem.

see:  `-1 AS inner_index,` or  `COALESCE(b.inner_index, -1) AS inner_index,`

desmond-hui (approved):

PR #369, opened 2023-10-19
update frequency and add test

tarikceric:
- Helius backfill has caught up so adjusting the run interval
- Running it twice a day, based on the assumption this table doesn't have to be updated too frequently. 
- The nft_mints table averages about 15k mints every 1/2 day, and the current job should get about 20k mints per run, so there's a decent buffer. If it does get backed up the test on the bronze_api__helius_nft_metadata should alert


desmond-hui (changes requested):
12 hour interval is fine but have it something other than midnight...we have many jobs that are kicking off at that time and causes a backlog

desmond-hui (approved):

PR #404, opened 2023-11-16
Hot fix/rewards backfill

tarikceric:
- add var for backfill job
- adjust epoch range (will revert back after backfilled)
- remove null tests for commission


models/silver/rewards/silver__rewards_rent.sql
```diff
@@ -100,12 +100,13 @@ epoch AS (
 
 {% if is_incremental() %}
 WHERE
-    epoch <= (
-        SELECT
-            MAX(epoch_earned) + 10
-        FROM
-            {{ this }}
-    )
+    epoch <= 540
```
desmond-hui, line 103: Please explain why this is hardcoded

.github/workflows/dbt_run_batch_backfill.yml
```diff
@@ -41,4 +42,4 @@ jobs:
           dbt deps
       - name: Run DBT Jobs
         run: |
-          dbt run -s models/silver/rewards/silver__rewards_fee.sql models/silver/rewards/silver__rewards_rent.sql models/silver/rewards/silver__rewards_voting.sql models/silver/rewards/silver__rewards_staking.sql --vars '{DBT_IS_BATCH_LOAD: true}'
+            dbt run -s models/silver/rewards/silver__rewards_rent.sql
```
desmond-hui, line 45: Please explain why we are not backfilling the other reward models

models/silver/rewards/silver__rewards_rent.sql
tarikceric: there were some spurts where the bronze table didn't contain rewards in epoch order, so sometimes it would be outside the +10 partition range and not grab anything -- the plan was to change it back after the backfill is complete

.github/workflows/dbt_run_batch_backfill.yml
tarikceric: The others are complete

desmond-hui (approved):

PR #414, opened 2023-11-29
fix for shares

eric-laurello:


models/silver/swaps/silver__swaps_intermediate_jupiterv5_1_view.sql
```diff
@@ -0,0 +1,8 @@
+{{ config(
+    materialized = 'view'
+) }}
+
+SELECT
+    *
+FROM
+    {{ ref('silver__swaps_intermediate_jupiterv6') }}
```
tarikceric, line 8: should be {{ source('solana_silver','swaps_intermediate_jupiterv5_1') }}

eric-laurello: oh good catch, copied the wrong line. will fix now

tarikceric (approved):

PR #469, opened 2024-01-26
ez_decoded---updates

eric-laurello:


models/silver/parser/silver__decoded_instructions.sql
```diff
@@ -4,7 +4,7 @@
     materialized = 'incremental',
     incremental_predicates = ["dynamic_block_date_ranges"],
     unique_key = "decoded_instructions_id",
-    cluster_by = ['block_timestamp::DATE','_inserted_timestamp::DATE','program_id'],
+    cluster_by = ['program_id','block_timestamp::DATE','_inserted_timestamp::DATE'],
```
desmond-hui, line 7: I have some reservations about putting program_id first in terms of incremental loading performance. Can we make this a V2 table for a week and create a separate GHA workflow to run this in parallel to see performance before we make the change?

models/gold/core/core__ez_events_decoded.sql
```diff
@@ -4,30 +4,30 @@
...
+    TRUE AS succeeded,
+    A.index,
     A.inner_index,
-    b.event_type,
-    b.program_id,
-    b.instruction,
+    A.event_type,
+    A.program_id,
+    NULL AS instruction,
     A.decoded_instruction,
```
desmond-hui, line 17: I think we should split this to be

```
A.decoded_instruction:accounts::array as decoded_accounts,
A.decoded_instruction:args::variant as decoded_args,
A.decoded_instruction:error::string as decoding_error
```

desmond-hui (changes requested):

models/gold/core/core__ez_events_decoded.sql
eric-laurello: @desmond-hui  Were you thinking about landing these so we can SO them or just breaking them out in the view is enough?

desmond-hui: No SO, i think view is enough

desmond-hui (approved):

PR #470, opened 2024-01-29
decoded_instructions_temp

eric-laurello:
Will run a clone to get the full table copied to test performance 1-1:
create Transient table solana_dev.silver.decoded_instructions_2  clone solana_dev.silver.decoded_instructions

### DIFF
--- a/models/silver/parser/silver__decoded_instructions_2.sql
+++ b/models/silver/parser/silver__decoded_instructions_2.sql
@@ -1,0 +1,91 @@
+-- depends_on: {{ ref('bronze__streamline_decoded_instructions_2') }}
+-- depends_on: {{ ref('bronze__streamline_FR_decoded_instructions_2') }}
+{{ config(
+    materialized = 'incremental',
+    incremental_predicates = ["dynamic_block_date_ranges"],
+    unique_key = "decoded_instructions_id",
+    cluster_by = ['program_id','block_timestamp::DATE','_inserted_timestamp::DATE'],
+    post_hook = enable_search_optimization(
+        '{{this.schema}}',
+        '{{this.identifier}}'
+    ),
+    merge_exclude_columns = ["inserted_timestamp"],
+    tags = ['scheduled_non_core'],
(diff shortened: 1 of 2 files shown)


models/silver/parser/silver__decoded_instructions_2.sql
```diff
@@ -0,0 +1,91 @@
...
+
+{% if is_incremental() %}
+{{ ref('bronze__streamline_decoded_instructions_2') }} A
+{% else %}
+    {{ ref('bronze__streamline_FR_decoded_instructions_2') }} A
+{% endif %}
+JOIN {{ ref('silver__blocks') }}
+b
+ON A.block_id = b.block_id
+JOIN {{ ref('silver__transactions') }} C
```
tarikceric, line 65: would it make sense to filter down the silver__transactions table to a date range as well? Thinking the join might be more efficient if its currently doing a larger scan of transactions

eric-laurello: Yeah was trying to think through how to do it, but a backfill could in theory pull in any dates right?

eric-laurello: what I could do is add the block_timestamp::DATE to the join

tarikceric: yea good point on the backfill, the range would be large either way. I'll make a separate ticket for improving the join

tarikceric (approved):

models/silver/parser/silver__decoded_instructions_2.sql
```diff
@@ -0,0 +1,92 @@
...
+ON A.tx_id = C.tx_id
+AND b.block_timestamp :: DATE = b.block_timestamp :: DATE
```
tarikceric, line 67: should the second be `c.block_timestamp`?

eric-laurello: good catch

PR #477, opened 2024-02-05
revert patch fix, force udf max batch rows, decrease batch size

desmond-hui:
- Force max batch size in UDF
  - This was causing some calls to pass multiple rows into the lambda instead of just one batch of requests. The lambda is currently only expecting 1 batch per call
- Decrease batch size
  - Saw some instances of longer running execution times, going to reduce it to give it more buffer
- Revert temp fix

### DIFF
--- a/models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
+++ b/models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
@@ -87,7 +87,7 @@ base AS (
 )
 SELECT
     ARRAY_AGG(request) AS batch_request,
-    solana_dev.streamline.udf_decode_compressed_mint_change_logs(batch_request) AS responses,
+    streamline.udf_decode_compressed_mint_change_logs(batch_request) AS responses,
     MIN(event_inserted_timestamp) AS start_inserted_timestamp,
     MAX(event_inserted_timestamp) AS end_inserted_timestamp,
     concat_ws(
--- a/models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
+++ b/models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
@@ -1,5 +1,4 @@
 -- depends_on: {{ ref('silver__nft_compressed_mints_onchain') }}
--- depends_on: {{ ref('silver__events') }}
 {{ config(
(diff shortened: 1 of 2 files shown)


models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
```diff
@@ -32,82 +31,71 @@
...
+        AND e._inserted_timestamp between '{{ min_inserted_timestamp }}' and '{{ max_inserted_timestamp }}'
+        AND NOT startswith(DATA, '2GJh')
+)
 SELECT
-    *
-FROM 
-    bronze_api.parse_compressed_nft_mints__final_tmp
-
+    ARRAY_AGG(request) AS batch_request,
+    solana_dev.streamline.udf_decode_compressed_mint_change_logs(batch_request) AS responses,
```
eric-laurello, line 90: do you want this pointing at dev?

models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
```diff
@@ -32,82 +31,71 @@
...
+                e._inserted_timestamp,
+                e.tx_id
+        ) AS rn,
+        FLOOR(
+            rn / 150
+        ) AS gn,
+        e._inserted_timestamp AS event_inserted_timestamp
+    FROM
+        collection_subset C
+        JOIN {{ ref('silver__events') }}
```
eric-laurello, line 74: maybe filtering events first in a cte would be faster? Not sure how long this takes to run now.

Only other comment is why have the block_timestamp in the where clause and not the join. Snowflake is probably  smart enough, but was wondering if you found it to perform better in the where .

models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
desmond-hui: Realised this, fixed it in latest commit

eric-laurello (approved):

models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
desmond-hui: No real reason, it behaves the same on the backend from what I can tell

PR #495, opened 2024-02-22
An 4547/solana stats

tarikceric:
create silver and gold table for general stats
- use `silver.blocks` for block-specific data because there are blocks that don't contain transactions


models/silver/stats/silver__core_metrics_hourly.sql
```diff
@@ -0,0 +1,117 @@
+{{ config(
+    materialized = 'incremental',
+    incremental_strategy = 'delete+insert',
+    unique_key = "block_timestamp_hour",
+    cluster_by = ['block_timestamp_hour::DATE'],
+    tags = ['curated']
```
eric-laurello, line 6: gotta update the tag here

tarikceric: added tag

eric-laurello (approved):

PR #496, opened 2024-02-27
An 4565/missing block txs and votes

desmond-hui:
- Add solscan gap test
- Add silver counts compared to solscan counts test
- Run new tests hourly
- Add new CTE to realtime block_txs view to re-ingest any data that comes from second test
  - This will self retry once per test run (every hour) if bronze data is successfully ingested (which doesnt guarantee it makes it to silver)
  - Limit amount we can retry to 200 for a single call


models/streamline/streamline__all_unknown_block_txs_real_time.sql
```diff
@@ -68,37 +83,22 @@ potential_missing_txs AS (
...
+        {{ source(
+            'solana_test_silver',
+            'transactions_and_votes_missing_7_days'
+        ) }}
+        m
+        LEFT JOIN {{ ref('streamline__complete_block_txs') }} C
+        ON C.block_id = m.block_id
+    WHERE
+        C._partition_id <= m._partition_id
+    LIMIT 200
```
eric-laurello, line 100: Is this intentional to make sure we don't overload the incremental?

desmond-hui: That's correct

eric-laurello (approved):

PR #502, opened 2024-03-01
fix stats and temp logic during block/tx fixes

tarikceric:


models/silver/stats/silver__core_metrics_hourly.sql
```diff
@@ -5,6 +5,38 @@
...
+    ) {% endset %}
+    {% set min_block_timestamp_hour = run_query(query).columns [0].values() [0]%}
+
+{% endif %}
+{% endif %}
+
+{% if execute %}
+{% if is_incremental() %}
+-- temp while older txs/blocks are being updated
+{% set query_1 = """ CREATE OR REPLACE TEMPORARY TABLE silver.core_metrics_hourly__intermediate_tmp AS SELECT distinct(block_timestamp)::date as dist_block FROM solana.silver.transactions WHERE _inserted_timestamp >= (SELECT MAX(_INSERTED_TIMESTAMP) FROM solana.silver.core_metrics_hourly)""" %}
```
eric-laurello, line 33: you should still be able to do the ref() in the query

models/silver/stats/silver__core_metrics_hourly.sql
```diff
@@ -30,13 +62,10 @@ WITH block_stats AS (
...
     'hour',
-    _inserted_timestamp
-) >= (
-    SELECT
-        MAX(DATE_TRUNC('hour', _inserted_timestamp)) - INTERVAL '12 hours'
-    FROM
-        {{ this }}
-)
+    block_timestamp
+) >= '{{ min_block_timestamp_hour }}'
```
eric-laurello, line 66: I think we can remove this where clause if we have the new one

eric-laurello (approved):

PR #505, opened 2024-03-08
add solsniper to nft sales

tarikceric:


models/silver/nfts/silver__nft_sales_solsniper.sql
```diff
@@ -0,0 +1,119 @@
...
+
+WITH base AS (
+
+    SELECT
+        *
+    FROM
+        {{ ref('silver__decoded_instructions_combined') }}
+    WHERE
+        program_id = 'SNPRohhBurQwrpwAptw1QYtpFdfEKitr4WSJ125cN1g'
+        AND event_type = 'executeSolNftOrder'
```
desmond-hui, line 17: We are missing NFT sales both regular and `cNFT` (not sure if latter is supposed to be included or not) as there are more than one NFT sales events.

```
select distinct event_type
from solana.silver.decoded_instructions_combined
where program_id = 'SNPRohhBurQwrpwAptw1QYtpFdfEKitr4WSJ125cN1g'
order by 1;
```

Example: `buyTensorPoolListingV1`

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: added the missing events, via a table+view since these events are deprecated

models/silver/nfts/silver__nft_sales_solsniper.sql
```diff
@@ -0,0 +1,136 @@
...
+
+{% if is_incremental() %}
+AND A._inserted_timestamp >= (
+    SELECT
+        MAX(_inserted_timestamp) - INTERVAL '1 day'
+    FROM
+        {{ this }}
+)
+{% else %}
+    AND A.block_timestamp :: DATE >= '2023-05-02'
```
desmond-hui, line 72: So, this is something across all our models using decoded data as our base...but I don't think it's appropriate to predicate on `_inserted_timestamp`

We are going to miss transfers due to backfills. I think we need to join based on `block_timestamp::date and tx_id`

I think we can/should generate the contiguous block_timestamp ranges of our decoded subset and use that predicate here

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_solsniper.sql
```diff
@@ -0,0 +1,136 @@
...
+        ON A.tx_id = b.tx_id
+        AND A.buyer_escrow_vault = b.tx_from
+        AND A.seller = b.tx_to
+        AND A.index = b.index_1
+        INNER JOIN transfers C
+        ON A.tx_id = C.tx_id
+        AND A.buyer_escrow_vault = C.tx_from
+        AND A.treasury_fee_account = C.tx_to
+        AND A.index = C.index_1
+    group by 1,2,3,4,5,6,7,8,9,10
```
desmond-hui, line 100: Also something to consider here...do we want to left join to transfers instead? What do we want to represent in the model?

- Do we want nft sales where we are able to create a "complete" record? AKA some nft sales will be missing in the output

OR

- Do we always want to capture what we deem as a sale via decoded data in the output? AKA some records may not have sales amount

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_solsniper.sql
desmond-hui: Also depending on your answer here...we may want to add a [`relationship` test](https://github.com/dbt-labs/dbt-utils?tab=readme-ov-file#relationships_where-source) in our upstream

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: logic updated

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: I like the second options approach of getting everything considered a sale. I was on the fence on whether something should be represented as a true sale if there are no transfers/sales amount, but I suppose if the program is identifying it as a sale it should be in there whether or not there was an exchange of tokens.

Updated to use left join. Not exactly sure how we want to do the relationship test

models/silver/nfts/silver__nft_sales_solsniper.sql
```diff
@@ -1,33 +1,48 @@
...
+    {% endset %}
+    {% set max_inserted_timestamp = run_query(max_inserted_query).columns [0].values() [0] %}
+{% endif %}
+
+{% set query = """ CREATE OR REPLACE TEMPORARY TABLE silver.decoded_instructions_solsniper__intermediate_tmp AS SELECT block_timestamp, block_id, tx_id, index, inner_index, program_id, decoded_instruction, event_type, _inserted_timestamp FROM """ ~ ref('silver__decoded_instructions_combined') ~ """ 
+    WHERE 
+        program_id = 'SNPRohhBurQwrpwAptw1QYtpFdfEKitr4WSJ125cN1g' 
+    AND 
+        event_type = 'executeSolNftOrder'""" %}     
+{% set incr = "" %}
```
desmond-hui, line 31: this is older code style, it would be better to make this a set so we don't need to do string concat 

Example
```
{% set abc %}
  create or replace table abc as 
  select * 
  from {{ ref('silver__events') }}
  limit 1
{% endset %}
{% do run_query(abc) %}
```

models/silver/nfts/silver__nft_sales_solsniper.sql
desmond-hui: I think you can start by trying to provided test from dbt_utils. Example below, unsure if `field` can be multiple keys like `tx_id, index, inner_index`

```
- dbt_utils.relationships_where:
              to: ref('nft_sales_solsniper')
              field: tx_id
              from_condition: program_id = 'sol_sniper_program_id' and event_type in ('my sales events') and _inserted_timestamp >= current_date - 2
              to_condition: _inserted_timestamp >= current_date - 2
```

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: Thanks for the great idea. This definitely seems like the appropriate check to make sure the solsniper model is returning everything identified as a sale. Updated the yml for decoded_instructions_combined

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: updated

models/silver/nfts/silver__nft_sales_solsniper.sql
```diff
@@ -23,23 +22,36 @@ FROM
...
+    block_id,
+    tx_id,
+    INDEX,
+    inner_index,
+    program_id,
+    decoded_instruction,
+    event_type,
+    _inserted_timestamp
+FROM
+    silver.decoded_instructions_combined
```
desmond-hui, line 38: you can use `ref` here

models/silver/nfts/silver__nft_sales_solsniper.sql
```diff
@@ -0,0 +1,165 @@
...
+    seller,
+    mint,
+    sales_amount,
+    _inserted_timestamp,
+    nft_sales_solsniper_id,
+    inserted_timestamp,
+    modified_timestamp,
+    invocation_id
+FROM
+    {{ ref('silver__nft_sales_solsniper_v1_events_view') }}
```
desmond-hui, line 165: I don't think we should be unioning a static view for every incremental load?

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: updated

models/silver/nfts/silver__nft_sales_solsniper.sql
tarikceric: updated to only run if not an incremental

desmond-hui: Won't the `_inserted_timestamp` be all weird? The max `_inserted_timestamp` will be the max of your incremental subset and this view after the first run. If the max on the view is > than the max of the subset then it wouldn't work?

I guess why have this view in here at all vs. only including it in `models/gold/nft/nft__fact_nft_sales.sql`?

desmond-hui (approved):

PR #516, opened 2024-03-14
new token2022 events

tarikceric:
Add new events from token 2022 standard program `TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb`

### DIFF
--- a/models/silver/silver__mint_actions.sql
+++ b/models/silver/silver__mint_actions.sql
@@ -22,7 +22,8 @@ WITH base_events AS (
     {% else %}
         AND _inserted_timestamp::date between '2022-08-12' and '2022-09-05'
     {% endif %}
-)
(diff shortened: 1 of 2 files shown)


models/silver/silver__mint_actions.sql
```diff
@@ -71,7 +75,8 @@ SELECT
     ) AS mint_amount,
     COALESCE(
         i.value :parsed :info :mintAuthority :: string,
-        i.value :parsed :info :multisigMintAuthority :: string
+        i.value :parsed :info :multisigMintAuthority :: string,
+        instruction :parsed :info :authority :: string
```
desmond-hui, line 79: Should this be `i.value:parsed :info :authority :: string`

PR #534, opened 2024-04-15
An 4746/fix c nft parsing

desmond-hui:
- Add CTE to filter out transactions that have already run through the compressed mints decoder
- Add onchain mint events for `F9SixdqdmEBP5kprp2gZPZNeMmfHJRCTMFjN22dx3akf` because it uses bubblegum
- Add txs from additional program `F9SixdqdmEBP5kprp2gZPZNeMmfHJRCTMFjN22dx3akf` to be decoded
- Increase workflow job calls by 2

### DIFF
--- a/models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
+++ b/models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
@@ -6,7 +6,7 @@
 ) }}
 
 {% if execute %}
-    {% set query = """
+    {% set query %}
         CREATE OR REPLACE TEMPORARY TABLE bronze_api.parse_compressed_nft_mints__intermediate_tmp AS 
         SELECT
             block_timestamp,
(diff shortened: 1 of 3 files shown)


models/bronze/bronze_api/bronze_api__parse_compressed_nft_mints.sql
```diff
@@ -6,35 +6,41 @@
...
+            FROM
+                {{ this }}
+        ),
+        base AS (
+            SELECT DISTINCT r.value:tx_id::string AS tx_id
+            FROM {{ this }}
+            JOIN TABLE(flatten(responses)) r
+            WHERE end_inserted_timestamp >= (SELECT max_inserted_timestamp FROM max_time)
+        )
         SELECT
```
tarikceric, line 23: the `base` cte @line 52 is returning duplicates due to the join so the parser is running the same data twice -- i think we can cut down on that by grabbing the `m.index`  here and adding `c.index = e.index` in the join

desmond-hui: I don't think I understand. The `base` CTE is grabbing `distinct tx_id` so there shouldn't be any duplicates. It is then used to exclude data for the next model run

tarikceric: was referring to the second base cte @ line 52 (wasn't able to put a comment there) -- was recommending grabbing an additional column in this first `base` cte, and then using it in the join in the `second` one

tarikceric (approved):

PR #537, opened 2024-04-16
update idl table

tarikceric:
- Include additional information in the gold IDL table. Based on user requests for more visibility on decoding process

### DIFF
--- a/models/gold/core/core__dim_idl.sql
+++ b/models/gold/core/core__dim_idl.sql
@@ -8,7 +8,12 @@ SELECT
     program_id,
     idl,
     idl_hash,
+    is_valid,
+    submitted_by,
+    _inserted_timestamp,
+    first_block_id,
     earliest_decoded_block,
+    backfill_status,
     idl_id as dim_idl_id,
     inserted_timestamp,
     modified_timestamp
--- a/models/gold/core/core__dim_idl.yml
+++ b/models/gold/core/core__dim_idl.yml
@@ -1,7 +1,7 @@
 version: 2
 models:
   - name: core__dim_idl
-    description: The status of Program IDL's submitted for decoding events 
+    description: Deprecating soon! This is a notice that we're replacing this table with `core__dim_idls`. Please migrate queries to the new table by 06/06/24.
     columns:
       - name: PROGRAM_ID
         description: "{{ doc('program_id') }}"
@@ -16,23 +16,7 @@ models:
         tests:
           - dbt_expectations.expect_column_to_exist
       - name: EARLIEST_DECODED_BLOCK
-        description: "{{ doc('earliest_decoded_block') }}"
-        tests:
-          - dbt_expectations.expect_column_to_exist
-      - name: IS_VALID
-        description: "{{ doc('is_valid') }}"
-        tests:
-          - dbt_expectations.expect_column_to_exist
-      - name: SUBMITTED_BY
-        description: "{{ doc('submitted_by') }}"
-        tests:
-          - dbt_expectations.expect_column_to_exist
-      - name: FIRST_BLOCK_ID
-        description: "{{ doc('first_block_id') }}"
-        tests:
-          - dbt_expectations.expect_column_to_exist
-      - name: BACKFILL_STATUS
-        description: "{{ doc('backfill_status') }}"
+        description: "The oldest block where program events has been decoded"
         tests:
           - dbt_expectations.expect_column_to_exist
       - name: DIM_IDL_ID
--- a/models/gold/core/core__dim_idl.yml
+++ b/models/gold/core/core__dim_idl.yml
@@ -16,7 +16,23 @@ models:
         tests:
           - dbt_expectations.expect_column_to_exist
       - name: EARLIEST_DECODED_BLOCK
-        description: "The oldest block where program events has been decoded"
+        description: "{{ doc('earliest_decoded_block') }}"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+      - name: IS_VALID
+        description: "{{ doc('is_valid') }}"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+      - name: SUBMITTED_BY
+        description: "{{ doc('submitted_by') }}"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+      - name: FIRST_BLOCK_ID
+        description: "{{ doc('first_block_id') }}"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+      - name: BACKFILL_STATUS
+        description: "{{ doc('backfill_status') }}"
         tests:
           - dbt_expectations.expect_column_to_exist
       - name: DIM_IDL_ID
--- a/models/silver/idls/silver__idl.sql
+++ b/models/silver/idls/silver__idl.sql
@@ -5,25 +5,8 @@
     tags = ['scheduled_non_core']
 ) }}
 
-WITH submitted_idls AS (
+WITH idls AS (
 
-    SELECT
-        A.program_id,
-        A.idl,
-        A.idl_hash,
-        A.is_valid,
-        A.discord_username,
-        A._inserted_timestamp,
-        b.first_block_id
-    FROM
-        {{ ref('silver__verified_user_idls') }} A
-        LEFT JOIN {{ ref('streamline__idls_history') }}
-        b
-        ON A.program_id = b.program_id qualify(ROW_NUMBER() over(PARTITION BY A.program_id
-    ORDER BY
-        A._inserted_timestamp DESC)) = 1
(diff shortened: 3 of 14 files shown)


models/silver/idls/silver__idl.sql
```diff
@@ -20,31 +37,68 @@ WHERE
...
+    SELECT
+        table_schema,
+        table_name,
+        SPLIT_PART(
+            table_name,
+            '_',
+            ARRAY_SIZE(SPLIT(table_name, '_'))
+        ) AS in_progress_program_id
+    FROM
+        solana_dev.information_schema.views
```
desmond-hui, line 50: Can remove `solana_dev`

models/silver/idls/silver__idl.sql
```diff
@@ -20,31 +37,68 @@ WHERE
...
+    is_valid,
+    discord_username as submitted_by,
+    _inserted_timestamp,
+    first_block_id,
+    earliest_decoded_block,
+    CASE
+        WHEN earliest_decoded_block = first_block_id THEN 'complete'
+        WHEN in_progress_program_id IS NOT NULL THEN 'in progress'
+        WHEN NOT is_valid THEN NULL
+        ELSE 'not started'
```
desmond-hui, line 96: Not sure @eric-laurello @tarikceric thoughts on this but I am leaning we should make these have `_` instead of space (i.e. `in_progress`)

models/silver/idls/silver__idl.sql
desmond-hui, line 1: I think the table should be named `silver.idls`? To be consistent with our other tables which are all plural AFAIK

Same for the core model...except not sure what change management process we want to use for that since it is existing

models/silver/idls/silver__idl.sql
```diff
@@ -20,31 +37,68 @@ WHERE
...
+        table_name,
+        SPLIT_PART(
+            table_name,
+            '_',
+            ARRAY_SIZE(SPLIT(table_name, '_'))
+        ) AS in_progress_program_id
+    FROM
+        solana_dev.information_schema.views
+    WHERE
+        table_name LIKE 'DECODED_INSTRUCTIONS_BACKFILL_%'
```
desmond-hui, line 52: @eric-laurello @tarikceric  Since we have two types of backfills, do we want to exclude retry backfills where we are mostly using it for when IDLs are updated?

models/silver/idls/silver__idl.sql
```diff
@@ -20,31 +37,68 @@ WHERE
...
+    ORDER BY
+        table_name DESC
```
desmond-hui, line 54: I don't think this order by is necessary

models/silver/idls/silver__idl.sql
```diff
@@ -20,31 +37,68 @@ WHERE
...
-LEFT JOIN {{ this }} C
-ON A.program_id = C.program_id
+LEFT JOIN {{ this }}
+d
+ON A.program_id = d.program_id
 WHERE
-    A.earliest_decoded_block < C.earliest_decoded_block
-    OR C.earliest_decoded_block IS NULL
+    b.earliest_decoded_block < d.earliest_decoded_block
+    OR d.earliest_decoded_block IS NULL
```
desmond-hui, line 80: Could potentially have some conflict here where the backfill status flag is changed but the earliest block is unchanged, the model would not pick up any other status or column changes. Also another example would be if an IDL is updated, the `idl_hash` would change but it probably would never be updated in this model due to this condition.

An inline `IFF` might work better here

models/silver/idls/silver__idl.yml
```diff
@@ -7,15 +7,20 @@ models:
         tests:
           - not_null
       - name: EARLIEST_DECODED_BLOCK
-        description: "The oldest block where program events has been decoded"
-        tests:
-          - not_null
+        description: "{{ doc('earliest_decoded_block') }}"
       - name: IDL
         description: "The complete IDL that defines the program"
+      - name: IDL_HASH
```
desmond-hui, line 13: Why was description removed here for `IDL_HASH`?

desmond-hui (changes requested):

models/silver/idls/silver__idl.sql
tarikceric: agreed that seems more standard

models/silver/idls/silver__idl.sql
tarikceric: will exclude RETRY backfills

models/silver/idls/silver__idl.sql
tarikceric: That makes sense, I'll update that after finalized so the comments arent lost from 'deleting' the existing table in the pr. I'll reach out to Liz on the communication for name changes

models/silver/idls/silver__idl.sql
tarikceric: ended up adding a union under the incremental, because i would want to use the existing 'earliest_block_decoded' if it just a idl_hash change or there's a backfill occurring but earliest block is unchanged -- what do you think of that approach?

models/silver/idls/silver__idl.yml
tarikceric: accidently removed it, good catch

models/silver/idls/silver__idl.sql
desmond-hui: That seems too complicated. I think you can use `iff` for the column itself and remove the WHERE clause and leave everything else the same as before

```
{% if is_incremental() %}
        iff(b.earliest_decoded_block < d.earliest_decoded_block, b.earliest_decoded_block, d.earliest_decoded_block) AS earliest_decoded_block,
        {% else %}
        b.earliest_decoded_block,
        {% endif %}
```

models/silver/idls/silver__idl.sql
```diff
@@ -20,31 +37,89 @@ WHERE
...
+    program_id,
+    idl,
+    idl_hash,
+    is_valid,
+    discord_username as submitted_by,
+    _inserted_timestamp,
+    first_block_id,
+    earliest_decoded_block,
+    CASE
+        WHEN earliest_decoded_block = first_block_id AND in_progress_program_id IS NOT NULL THEN 'complete'
```
desmond-hui, line 114: Why are we checking for `in_progress_program_id IS NOT NULL` here? Wouldn't `in_progress_program_id` be NULL when the backfill is `complete`?

desmond-hui (changes requested):

models/silver/idls/silver__idl.sql
desmond-hui: That seems too complicated. I think you can use `iff` for the column itself and remove the WHERE clause and leave everything else the same as before

```
{% if is_incremental() %}
        iff(b.earliest_decoded_block < d.earliest_decoded_block, b.earliest_decoded_block, d.earliest_decoded_block) AS earliest_decoded_block,
        {% else %}
        b.earliest_decoded_block,
        {% endif %}
```

models/gold/core/core__dim_idl.sql
```diff
@@ -8,7 +8,12 @@ SELECT
     program_id,
     idl,
     idl_hash,
+    is_valid,
+    submitted_by,
+    _inserted_timestamp,
```
desmond-hui, line 13: Do we want to expose `_inserted_timestamp` in gold layer?

models/silver/idls/silver__idl.sql
tarikceric: sounds good, updated to use that approach

models/silver/idls/silver__idl.sql
tarikceric: My mistake, I was thinking about when we do retry backfills but we're excluding that

models/gold/core/core__dim_idl.sql
desmond-hui: Lets rename it

models/gold/core/core__dim_idl.yml
```diff
@@ -16,7 +16,23 @@ models:
...
+      - name: SUBMITTED_BY
+        description: "{{ doc('submitted_by') }}"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+      - name: FIRST_BLOCK_ID
+        description: "{{ doc('first_block_id') }}"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+      - name: BACKFILL_STATUS
+        description: "{{ doc('backfill_status') }}"
```
desmond-hui, line 35: Add description for `date_submitted`

models/gold/core/core__dim_idl.sql
desmond-hui: Lets rename it

desmond-hui (approved):

PR #540, opened 2024-04-23
An 4769/me solsniper cnft sales

tarikceric:
cNFT sales for solsniper/magic eden
- first get all sales via `_onchain` models, then decode mint via `bronze_api` models, and join into final table


models/silver/nfts/silver__nft_sales_solsniper_cnft_onchain.yml
```diff
@@ -0,0 +1,60 @@
...
+              datepart: day
+              interval: 2
+      - name: BLOCK_ID
+        description: "{{ doc('block_id') }}"
+        tests:
+          - not_null
+      - name: TX_ID
+        description: "{{ doc('tx_id') }}"
+        tests:
+          - not_null
```
desmond-hui, line 24: Add the included or custom relationship test to make sure all our onchain events make it to the destination table

models/silver/nfts/silver__nft_sales_solsniper_cnft.sql
```diff
@@ -0,0 +1,70 @@
...
+        ['a.tx_id','b.mint']
+    ) }} AS nft_sales_solsniper_cnft_id,
+    SYSDATE() AS inserted_timestamp,
+    SYSDATE() AS modified_timestamp
+FROM
+    {{ ref('silver__nft_sales_solsniper_cnft_onchain') }} A
+    LEFT JOIN mint_addresses b
+    ON A.tx_id = b.tx_id
+WHERE
+    b.mint IS NOT NULL
```
desmond-hui, line 61: I think we should include all the sales, even if we don't have the mint. Add a test to check for NULL mints so we are alerted and can triage the pipeline

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_solsniper_cnft_onchain.yml
tarikceric: added relationship test for solsniper and magic_eden_cnft_onchain to decoded_instructions_combined -- or were you thinking a relationship between the `_cnft_onchain` and final `_cnft` table?

models/silver/nfts/silver__nft_sales_solsniper_cnft.sql
tarikceric: removed filter from both models. I think the `not_null` test for the `mint` column in the yml should suffice as a check

models/silver/nfts/silver__nft_sales_magic_eden_cnft.sql
desmond-hui, line 1: How would this (and the solsniper model) "auto-heal"? Meaning, at this moment we now have N records w/ NULL mints because we are waiting for a backfill of the bronze data. How do we fix the NULL records as the bronze data arrives?

desmond-hui (changes requested):

models/silver/nfts/silver__nft_sales_magic_eden_cnft.sql
tarikceric: great point -- added temp backfill logic using an _inserted_timestamp from bronze table, using a temp `mint_inserted_timestamp` column to track the latest records that have went through the parser

desmond-hui (approved):

PR #576, opened 2024-06-11
Add b58 decoding to logs extraction udf

desmond-hui:
- Handle additional log extraction edge cases
  - Handle logs that end in `" success"` but are not actually program log end
  - Handle logs being truncated as last message
  - Handle logs being truncated but additional messages are rendered before it ends
  - Handle legacy BPG program log format
- Ignore txs with `NULL` log_messages in model
- Add `_udf_error` column to schema
- Fix missing doc ref comment for logs realtime query

### DIFF
--- a/models/silver/program_logs/silver__transaction_logs_program_data.yml
+++ b/models/silver/program_logs/silver__transaction_logs_program_data.yml
@@ -44,6 +44,9 @@ models:
         description: "{{ doc('program_id') }}"
         tests: 
           - not_null: *recent_date_filter
+      - name: _UDF_ERROR
+        description: >
+          Error from `udf_get_logs_program_data` if unsuccessful
       - name: _INSERTED_TIMESTAMP
         description: "{{ doc('_inserted_timestamp') }}"
         tests: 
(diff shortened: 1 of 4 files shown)


models/silver/program_logs/silver__transaction_logs_program_data.yml
```diff
@@ -44,6 +44,9 @@ models:
         description: "{{ doc('program_id') }}"
         tests: 
           - not_null: *recent_date_filter
+      - name: _UDF_ERROR
```
tarikceric, line 47: would a test here be helpful, like if its over a % threshold it would alert?

tarikceric (approved):

models/silver/program_logs/silver__transaction_logs_program_data.yml
desmond-hui: The existing null threshold test only works the other way, i.e for columns that should be mostly non-null. However, I added `dbt_expectations.expect_column_values_to_be_null` with no threshold, we will see how that goes, currently passes.

```
14:00:09  1 of 15 PASS dbt_expectations_expect_column_values_to_be_null_silver__transaction_logs_program_data__UDF_ERROR___inserted_timestamp_current_date_7  [PASS in 2.84s]
```

PR #582, opened 2024-06-12
ez_dex_swaps

tarikceric:
Create defi.ez_dex_swaps view
- following naming conventions present in fact_swaps
- no raw amounts as those are not tracked in Solana tables

### DIFF
--- a/models/gold/defi/defi__ez_dex_swaps.sql
+++ b/models/gold/defi/defi__ez_dex_swaps.sql
@@ -25,6 +25,15 @@ WITH swaps AS (
         {{ ref('defi__fact_swaps') }}
     WHERE
         succeeded
+),
+prices AS (
+    SELECT
+        HOUR,
+        token_address,
+        symbol,
+        price
+    FROM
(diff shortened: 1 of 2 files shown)


models/gold/defi/defi__ez_dex_swaps.sql
```diff
@@ -0,0 +1,67 @@
...
+    d.ez_swaps_id,
+FROM
+    swaps d
+    LEFT JOIN {{ ref('price__ez_prices_hourly') }} p_in
+    ON d.swap_from_mint = p_in.token_address
+    AND DATE_TRUNC(
+        'hour',
+        d.block_timestamp
+    ) = p_in.hour
+    LEFT JOIN solana.price.ez_prices_hourly p_out
```
eric-laurello, line 62: two small things:

1. remove the hard coding
2. put prices in a cte so it only needs to be ref'd once

tarikceric: updated with a cte for prices

eric-laurello (approved):

PR #591, opened 2024-06-24
recency test on inserted_timestamp

tarikceric:
change recency test to be on inserted_timestamp. New records won't always be grabbed during the runs but we expect the runs to update the inserted_timestamp each time.

### DIFF
--- a/models/silver/validator/silver__snapshot_stake_accounts.yml
+++ b/models/silver/validator/silver__snapshot_stake_accounts.yml
@@ -42,6 +42,9 @@ models:
         description: "{{ doc('_inserted_timestamp') }}"
         tests:
           - not_null
+      - name: INSERTED_TIMESTAMP
+        description: '{{ doc("inserted_timestamp") }}'   
+        tests: 
           - dbt_expectations.expect_row_values_to_have_recent_data:
               datepart: hour
               interval: 12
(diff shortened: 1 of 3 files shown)


desmond-hui (commented):
Overall concern with this change is if the model inserted the same data over and over again, this would never fail but I think it **should** fail "recency". We may be mean two things with the same word. One form of recency is "was there recent data incrementally loaded into this model" and the other form is "was the model recently run successfully". I think this form only tests the latter but not the former. Is there an expectation for how recent our data should be?

tarikceric:
You're right that this test only really checks if ' the model recently run successfully'. We expect new data every epoch, so ~2.5 days, but the issue is if we don't get anything in that window then we can't retrieve it since those calls only return data for the current epoch. During an epoch it would insert the same data again, and there shouldn't be any changes (besides a very small percentage of stake accts that go through merges/splits) -- my thought was this would address that issue of confirming the API calls are working properly, and then we can fix it before the next epoch starts.

desmond-hui:
If it is ~2.5 days then can we also have recency test on `_inserted_timestamp` of 3 or 4 days or would that not be useful?

tarikceric:
I dont think it would be useful. The `inserted_timestamp` test would alert before the one for `_inserted_timestamp`, and also by the time the '_inserted_timestamp' alerts we'd be out of luck because the epoch is passed.

desmond-hui (approved):

PR #596, opened 2024-07-02
Jup gold views

desmond-hui:
- Add new gold layer models for jupiter inner and summary swaps
- Add refactored jupiterv6 summary swap table into `fact_swaps` for backwards compatability
  - Aggregator swaps should be removed from `fact_swaps` at some point in the future
- Disable jupiterv6 version 1 model
  - Add as source and view but should no longer be updated
- Change `fact_swaps` test filter to take advantage of partition pruning 

```
-- rebuild gold views on M
-- dbt build -s models/silver/swaps/jupiter/v6/silver__swaps_intermediate_jupiterv6_view.sql models/gold/defi/defi__fact_swaps_jupiter_inner.sql models/gold/defi/defi__fact_swaps_jupiter_summary.sql models/gold/defi/defi__fact_swaps.sql

14:43:41  Finished running 4 view models, 45 tests, 9 hooks in 0 hours 6 minutes and 41.18 seconds (401.18s).
```


models/gold/defi/defi__fact_swaps.sql
```diff
@@ -50,16 +51,43 @@ SELECT
     to_mint AS swap_to_mint,
     program_id,
     l.address_name AS swap_program,
-    concat_ws('-',tx_id,swap_index) as _log_id,
+    concat_ws('-',tx_id,swap_index,swap_program) as _log_id,
```
desmond-hui, line 54: @tarikceric - Should we deprecate `_log_id` column?

tarikceric: Its still being used in the crosschain swaps table so I think it should be left in

tarikceric (approved):

PR #604, opened 2024-07-12
refactor incremental logic

desmond-hui:
- Fix issue with `OR` statement in `sellers` CTE where it was causing full event scans when older blocks were re-processed
- Add `between_stmts` filtering for applicable CTEs
- Use `get_batch_load_logic` macro instead of old way of batch loading for backfills
  - No backfill needed here, just cleans up the code a bit

### DIFF
--- a/models/silver/nfts/silver__nft_sales_magic_eden_v2.sql
+++ b/models/silver/nfts/silver__nft_sales_magic_eden_v2.sql
@@ -6,62 +6,43 @@
     tags = ['scheduled_non_core']
 ) }}
 
-WITH txs AS (
+/* run incremental timestamp value first then use it as a static value */
+{% if execute %}
+    {% set create_tmp_query %}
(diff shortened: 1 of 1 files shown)


models/silver/nfts/silver__nft_sales_magic_eden_v2.sql
```diff
@@ -232,43 +161,15 @@ post_token_balances AS (
...
-        FROM
-            {{ this }}
-    )
-{% else %}
-WHERE
-    _inserted_timestamp :: DATE BETWEEN '2022-08-01'
-        AND '2022-09-01'
-{% endif %}
+    WHERE
+        1=1
```
tarikceric, line 165: should there be a `AND {{ between_stmts }}`?

desmond-hui: Don't think that's necessary. It's already filtering by `_inserted_timestamp >= max()`. I added for `events` because that was the model causing the slow down based on the query triage. Specifically the `sellers` cte.

tarikceric (approved):

PR #617, opened 2024-07-23
Hotfix solana txs votes

desmond-hui:
- Modify `silver.transactions` logic to handle txs that could have non-vote/non-compute instructions within a vote transaction

To reconcile I ran a few partitions through both the existing and new model logic. These vote txs with non-vote instructions are rare so the expectation would be for these 2 versions to produce no diffs of the new one to have slightly more (1 to 2 txs). In the real-time samples I tested, they produced no diffs. In the partition where I know one of these existed, they produced the expected diff.

Screenshot 2024-07-23 at 8 37 10 AM
Screenshot 2024-07-23 at 8 43 05 AM
Screenshot 2024-07-23 at 8 44 29 AM

Also there does not seem to be any noticeable performance degradation for runtime (expected 1-2mins)

Screenshot 2024-07-23 at 8 38 58 AM

### DIFF
--- a/models/silver/silver__transactions.sql
+++ b/models/silver/silver__transactions.sql
@@ -43,10 +43,13 @@ WITH pre_final AS (
         ON b.block_id = t.block_id
     WHERE
         tx_id IS NOT NULL
-        AND COALESCE(
-            t.data :transaction :message :instructions [0] :programId :: STRING,
-            ''
-        ) <> 'Vote111111111111111111111111111111111111111'
+        AND (
+            COALESCE(t.data :transaction :message :instructions [0] :programId :: STRING,'') <> 'Vote111111111111111111111111111111111111111'
+            OR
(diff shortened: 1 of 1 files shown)


models/silver/silver__transactions.sql
```diff
@@ -112,8 +114,35 @@ prev_null_block_timestamp_txs AS (
     WHERE
         t.block_timestamp :: DATE IS NULL
         AND t.block_id > 39824213
-)
+),
 {% endif %}
+qualifying_transactions AS (
+    SELECT
+        tx_id,
+        array_agg(i.value:programId::string) within group (order by i.index) as program_ids
```
eric-laurello, line 122: any reason why to just not do a count here? Are there transactions without instructions?

eric-laurello (approved):

PR #625, opened 2024-07-29
Fix/solscan test for new silver transactions logic

desmond-hui:
- Adjust `bronze` vs. `silver` transaction count comparison tests to account for the fact that `silver.transactions` and `silver.votes` can now have the same `tx_id` in both models
  - There will be a performance degradation as we now have to use `UNION` instead of `UNION ALL`. Should still be fine since these only run once per hour

Tests pass against prod w/ changed logic
```
(dbt-env) dbt test -s test_silver__transactions_partial_partitions -t prod
14:42:09  1 of 1 PASS test_silver__transactions_partial_partitions ....................... [PASS in 74.43s]

(dbt-env) dbt test -s test_silver__transactions_and_votes_missing_7_days -t prod
14:55:19  1 of 1 PASS test_silver__transactions_and_votes_missing_7_days ................. [PASS in 855.73s]
```

### DIFF
--- a/tests/test_silver__transactions_partial_partitions.sql
+++ b/tests/test_silver__transactions_partial_partitions.sql
@@ -31,8 +31,6 @@ WITH max_loaded_part as (
         {{ ref('silver__transactions') }}
     WHERE
         _partition_id between (select max_partition_id-10 from max_loaded_part) and (select max_partition_id from max_loaded_part)
-    GROUP BY
-        1,2,3
     UNION
     SELECT
         block_id,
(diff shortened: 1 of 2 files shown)


tests/test_silver__transactions_partial_partitions.sql
```diff
@@ -16,40 +16,38 @@ WITH max_loaded_part as (
...
     SELECT
         block_id,
         _partition_id,
-        COUNT(*) AS cnt
+        tx_id
     FROM
         {{ ref('silver__transactions') }}
     WHERE
         _partition_id between (select max_partition_id-10 from max_loaded_part) and (select max_partition_id from max_loaded_part)
     GROUP BY
```
eric-laurello, line 34: can kill the group by here and line 49

eric-laurello (approved):

PR #636, opened 2024-08-14
An 5088/gold optimization 2

tarikceric:
Search optimization and clustering for NFT and GOV schema

1. For larger/complex core views with union/join logic, convert to table and update silver clustering
- add logic to select from legacy views only during FR
- includes: `gov.fact_stake_accounts`, `gov.fact_validators`, `gov.fact_vote_accounts`, `nft.fact_nft_sales`)

3. For views with simpler logic, add SO and clustering to silver references

FR successful for core tables (on 2xl):
```
-- 23:47:48  2 of 5 OK created sql incremental model gov.fact_validators .................... [SUCCESS 1 in 23.25s]
-- 23:47:53  3 of 5 OK created sql incremental model gov.fact_vote_accounts ................. [SUCCESS 1 in 28.33s]
-- 23:48:27  4 of 5 OK created sql incremental model nft.fact_nft_sales ..................... [SUCCESS 1 in 61.76s]
-- 23:48:37  1 of 5 OK created sql incremental model gov.fact_stake_accounts ................ [SUCCESS 1 in 71.95s]
```
Incremental runs for core tables (on medium):
```
02:59:22  3 of 5 OK created sql incremental model gov.fact_vote_accounts ................. [SUCCESS 6300 in 7.92s]
02:59:23  4 of 5 OK created sql incremental model nft.fact_nft_sales ..................... [SUCCESS 266 in 8.62s]
02:59:24  2 of 5 OK created sql incremental model gov.fact_validators .................... [SUCCESS 1512 in 9.75s]
02:59:35  1 of 5 OK created sql incremental model gov.fact_stake_accounts ................ [SUCCESS 1442330 in 20.77s]
```

### DIFF
--- a/models/gold/gov/gov__fact_validators.sql
+++ b/models/gold/gov/gov__fact_validators.sql
@@ -5,7 +5,7 @@
     cluster_by = ['epoch','epoch_active'],
     merge_exclude_columns = ["inserted_timestamp"],
     post_hook = enable_search_optimization('{{this.schema}}','{{this.identifier}}','ON EQUALITY(node_pubkey, vote_pubkey)'),
-    tags = ['scheduled_non_core','exclude_change_tracking']
+    tags = ['scheduled_non_core']
 ) }}
 
 {% if execute %}
--- a/models/gold/gov/gov__fact_validators.sql
+++ b/models/gold/gov/gov__fact_validators.sql
@@ -1,8 +1,23 @@
 {{ config(
-  materialized = 'view',
-  meta ={ 'database_tags':{ 'table':{ 'PURPOSE': 'VALIDATOR' }} },
-  tags = ['scheduled_non_core','exclude_change_tracking']
+    materialized = 'incremental',
+    meta ={ 'database_tags':{ 'table':{ 'PURPOSE': 'VALIDATOR' }}},
+    unique_key = ['fact_validators_id'],
+    cluster_by = ['epoch','epoch_active'],
+    merge_exclude_columns = ["inserted_timestamp"],
+    post_hook = enable_search_optimization('{{this.schema}}','{{this.identifier}}','ON EQUALITY(node_pubkey, vote_pubkey)'),
+    tags = ['scheduled_non_core','exclude_change_tracking']
 ) }}
+
+{% if execute %}
+    {% if is_incremental() %}
+        {% set query %}
+            SELECT MAX(modified_timestamp) AS max_modified_timestamp
+            FROM {{ this }}
+        {% endset %}
+
+        {% set max_modified_timestamp = run_query(query).columns[0].values()[0] %}
+    {% endif %}
+{% endif %}
 
 SELECT
   epoch_recorded :: INT AS epoch,
@@ -42,6 +57,11 @@ SELECT
   ) AS modified_timestamp
 FROM
   {{ ref('silver__snapshot_validators_app_data') }}
+{% if is_incremental() %}
+WHERE
+    modified_timestamp >= '{{ max_modified_timestamp }}'
+{% endif %}
+{% if not is_incremental() %}
 UNION ALL
 SELECT
   epoch_recorded :: INT AS epoch,
@@ -72,3 +92,4 @@ SELECT
   '2000-01-01' AS modified_timestamp
 FROM
   {{ ref('silver__historical_validator_app_data') }}
+{% endif %}
--- a/models/gold/nft/nft__fact_nft_sales.sql
+++ b/models/gold/nft/nft__fact_nft_sales.sql
@@ -275,7 +275,7 @@ SELECT
     COALESCE (
         nft_sales_solanart_id,
         {{ dbt_utils.generate_surrogate_key(
-            ['tx_id']
+            ['tx_id','mint']
         ) }}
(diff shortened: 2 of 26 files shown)


models/gold/gov/gov__fact_validators.sql
```diff
@@ -1,9 +1,24 @@
...
-  materialized = 'view',
-  meta ={ 'database_tags':{ 'table':{ 'PURPOSE': 'VALIDATOR' }} },
-  tags = ['scheduled_non_core','exclude_change_tracking']
+    materialized = 'incremental',
+    meta ={ 'database_tags':{ 'table':{ 'PURPOSE': 'VALIDATOR' }}},
+    unique_key = ['fact_validators_id'],
+    cluster_by = ['epoch','epoch_active'],
+    merge_exclude_columns = ["inserted_timestamp"],
+    post_hook = enable_search_optimization('{{this.schema}}','{{this.identifier}}','ON EQUALITY(node_pubkey, vote_pubkey)'),
+    tags = ['scheduled_non_core','exclude_change_tracking']
```
desmond-hui, line 8: I think this only excluded change tracking because it was a view with join/union. We should be able to enable it now that it is a table

models/gold/nft/nft__fact_nft_sales.sql
```diff
@@ -1,9 +1,27 @@
 {{ config(
-    materialized = 'view',
-    meta ={ 'database_tags':{ 'table':{ 'PURPOSE': 'NFT' }} },
+    materialized = 'incremental',
+    meta ={ 'database_tags':{ 'table':{ 'PURPOSE': 'NFT' }}},
+    unique_key = ['fact_nft_sales_id'],
+    incremental_predicates = ["dynamic_range_predicate", "block_timestamp::date"],
+    cluster_by = ['block_timestamp::DATE','marketplace','is_compressed','program_id'],
```
desmond-hui, line 6: seems...odd that we woudl have both marketplace and program_id in the cluster key. Arent they are 1 to 1 relation?

models/gold/nft/nft__fact_nft_sales.sql
```diff
@@ -133,17 +134,15 @@ SELECT
     NULL as leaf_index,
     FALSE as is_compressed,
     {{ dbt_utils.generate_surrogate_key(
-        ['tx_id', 'mint']
+        ['tx_id']
```
desmond-hui, line 137: Reason that some of these are `tx_id` some are `[tx_id, mint]`?

desmond-hui (changes requested):

models/gold/nft/nft__fact_nft_sales.sql
tarikceric: Yup youre correct, but was assuming a user might query by `marketplace` instead of `program_id` so added it in. Is it not necessary then?

models/gold/nft/nft__fact_nft_sales.sql
tarikceric: Just based off our existing logic for the view and silver tables so assume that's the unique id that was decided for when the model was made

ie. There are silver nft_sales_xxx models that are just tx_id (ie silver__nft_sales_magic_eden_v2) so used that in the view. 

But good point bc there are models like solanart where silver uses tx_id + mint but the view only used tx_id.

Should i make the view column consistent with what the corresponding silver models are using for the _id? or adjust the ones only using tx_id to also include mint? Theres then also tensorswap which uses `['tx_id','mint','purchaser']`

models/gold/nft/nft__fact_nft_sales.sql
desmond-hui: I am just not sure it's worth clustering by 2 keys that are essentially the same...probably put one in SO instead?

models/gold/nft/nft__fact_nft_sales.sql
desmond-hui: Make the view consistent with the corresponding silver model. Are there any **active** marketplaces where we are only using `tx_id`?

models/gold/nft/nft__fact_nft_sales.sql
tarikceric: yea that makes sense to me, moved marketplace to SO

models/gold/nft/nft__fact_nft_sales.sql
tarikceric: `silver__nft_sales_magic_eden_v2` is the one active model that only uses tx_id

desmond-hui: Whats the effort to change that to `tx_id, mint`? I feel like it should be more specific than just tx_id. Are there really no txs with more than one NFT sale for ME2?

tarikceric: I suppose could do an ad-hoc query to update that col to use `tx_id+mint`, and just update the silver model logic. It seems like it really is just one nft sale per tx 

`select count(distinct(tx_id)),count(*) from solana.silver.nft_sales_magic_eden_v2;`

desmond-hui (approved):

PR #667, opened 2024-10-07
get all distinct txs; null amt/mint bug

tarikceric:
Handle edge case where the action is labeled as a swap but the user only transfers a mint but does not receive one.
- example https://solscan.io/tx/5KboaM89D79KwHwmRaoaSA7wD3xduMpdqVnWWxkEYvi4b16dTKBQBtGQYoQ1PgStD66mF95vi7FGNrcfmW3EGTpB
- Join jupiterv6 table as this only affects meteora swaps with a route, the heavy lifting for the logic has been done in the jup model, and show consistent values across the tables

Logic also added to base query as I noticed issues where decoded data and the interval result in incorrect swap values
- ie tx `42a7GeWqCVi8p3e5EGqKh6pcYnuqK4c6Kjg9ZmiN7SKe1aFhtvUDQJrPiMinTwvKPeAcTVVYjXE4TH33mMH52yD5`

Test
```
22:20:58  Finished running 15 data tests, 11 project hooks in 0 hours 1 minutes and 17.13 seconds (77.13s).
22:20:58  
22:20:58  Completed successfully
22:20:58  
22:20:58  Done. PASS=15 WARN=0 ERROR=0 SKIP=0 TOTAL=15
```

incremental
`23:07:18  1 of 1 OK created sql incremental model silver.swaps_intermediate_meteora ...... [SUCCESS 39465 in 59.56s]`


models/silver/swaps/meteora/silver__swaps_intermediate_meteora.sql
```diff
@@ -245,18 +265,63 @@ pre_final AS (
         ROW_NUMBER() over (PARTITION BY A.tx_id, A.index, A.inner_INDEX
             ORDER BY inner_index
         ) = 1
+),
+jup_decoded_swaps as (
```
desmond-hui, line 269: Is there no way to model this without relying on decoded jupiter data?

models/silver/swaps/meteora/silver__swaps_intermediate_meteora.sql
```diff
@@ -245,18 +265,63 @@ pre_final AS (
...
+        ) AS to_amt,
+        COALESCE(
+            p.to_mint,
+            j.to_mint
+        ) AS to_mint,
+        p._inserted_timestamp
+FROM
+    pre_final p
+    LEFT JOIN jup_decoded_swaps j
+    ON p.tx_id = j.tx_id
```
desmond-hui, line 311: nit: formatting

desmond-hui (approved):

models/silver/swaps/meteora/silver__swaps_intermediate_meteora.sql
tarikceric: There werent any identifiable differences in the decoded data between these edge cases + regular swaps, so couldn't use any clean logic and would have to use mints which IMO would overcomplicate the code for such few edge cases. Also wanted to remain consistent with what we have in jupiter.

PR #669, opened 2024-10-09
An 5047/helius cnft metadata

desmond-hui:
- Add streamline 2.0 pipeline for getting cNFT metadata from Helius
  - Calls the batch endpoint that allows us to grab 1000 nft metadata per call. We call in batch sizes of 10 w/ producer size of 100, so each run should get 100k nft metadata records
  - This will start from the very first cnft record by `_inserted_timestamp` in `silver__nft_compressed_mints`

Data in DEV
```
select *
from solana_dev.bronze.streamline_helius_nft_metadata
limit 100;

select *
from solana_dev.streamline.complete_helius_nft_metadata_requests;
```

`dbt run --vars '{"UPDATE_UDFS_AND_SPS": True}' -s bronze__streamline_helius_cnft_metadata bronze__streamline_FR_helius_cnft_metadata -t dev`
```
14:17:52  2 of 2 OK created sql view model bronze.streamline_helius_cnft_metadata ........ [SUCCESS 1 in 1.89s]
14:17:53  1 of 2 OK created sql view model bronze.streamline_FR_helius_cnft_metadata ..... [SUCCESS 1 in 2.21s]
```

`dbt run -s streamline__complete_helius_cnft_metadata_requests -t dev`
```
14:20:49  1 of 1 OK created sql incremental model streamline.complete_helius_cnft_metadata_requests  [SUCCESS 1 in 121.06s]
```

`dbt run --vars '{STREAMLINE_INVOKE_STREAMS: True}' -s streamline__helius_cnft_metadata_requests -t dev`
```
14:22:34  Running macro `if_data_call_function`: Calling udf streamline.udf_bulk_rest_api_v2 with params: 
{
  "exploded_key": "[\"result\"]",
  "external_table": "helius_nft_metadata",
  "order_by_column": "group_num",
  "producer_batch_size": "100",
  "sql_limit": "100",
  "worker_batch_size": "10"
}
14:23:40  1 of 1 OK created sql view model streamline.helius_cnft_metadata_requests ...... [SUCCESS 1 in 65.19s]
```


models/streamline/nft/metadata/requests/streamline__helius_cnft_metadata_requests.sql
```diff
@@ -0,0 +1,101 @@
...
+    FROM
+        {{ ref('silver__nft_compressed_mints') }}
+    WHERE 
+        _inserted_timestamp >= (
+            SELECT 
+                coalesce(max(max_mint_event_inserted_timestamp), '2000-01-01') 
+            FROM 
+                {{ ref('streamline__complete_helius_cnft_metadata_requests') }}
+        )
+    EXCEPT
```
tarikceric, line 32: wondering if this can be filtered more, since the size of complete_helius_cnft_metadata_requests is going to be pretty large so runtime will keep increasing

desmond-hui: So we are comparing A to B. A = `silver__nft_compressed_mints` and B = `streamline__complete_helius_cnft_metadata_requests`. A is being reduced on each subsequent run, meaning there is less to compare and thus should yield faster results. There is no good way to filter on B, you need to know what has been done but I think if you limit A, the size of B will not matter as much.

Example, compare these 2. B is constant, A is reduced in the 2nd comparison and it runs much faster

```
select block_id
from solana.silver.blocks
except
select block_id
from solana.silver.blocks;

select block_id
from (
    select block_id
    from solana.silver.blocks
    limit 1000
)
except
select block_id
from solana.silver.blocks;
```

tarikceric (approved):

PR #678, opened 2024-10-17
An 5047/helius silver and gold models

desmond-hui:
- Parse out metadata from helius raw bronze data for cnfts

FR on 2xl w/ prod data
```
14:33:20  1 of 9 OK created sql incremental model silver.helius_cnft_metadata ............ [SUCCESS 1 in 882.45s]
```

Incremental on M w/ prod data
```
14:35:29  1 of 9 OK created sql incremental model silver.helius_cnft_metadata ............ [SUCCESS 140000 in 52.57s]
```

Gold view
```
14:43:53  1 of 1 OK created sql view model nft.dim_nft_metadata .......................... [SUCCESS 1 in 2.30s]
```


models/silver/metadata/helius/silver__helius_cnft_metadata.yml
```diff
@@ -0,0 +1,60 @@
...
+        tests:
+          - not_null:
+              config:
+                where: >
+                  helius_collection_id is not null
+                  AND _inserted_timestamp::date >= current_date - 7
+      - name: _INSERTED_TIMESTAMP
+        description: "{{ doc('_inserted_timestamp') }}"
+        tests: 
+          - not_null: *recent_date_filter
```
tarikceric, line 41: filter wouldnt work here?

models/silver/metadata/helius/silver__helius_cnft_metadata.yml
```diff
@@ -0,0 +1,60 @@
...
+        description: "{{ doc('image_url') }}"
+      - name: METADATA_URI
+        description: "{{ doc('token_metadata_uri') }}"
+      - name: NFT_NAME
+        description: "The name of the NFT"
+      - name: HELIUS_COLLECTION_ID
+        description: >
+          The ID of the collection from Helius
+      - name: NFT_COLLECTION_ID
+        description: >
```
tarikceric, line 30: nit: any difference between using `>` vs the desc in quotes?

tarikceric (approved):

models/silver/metadata/helius/silver__helius_cnft_metadata.yml
desmond-hui: So `>` symbol is used for single line string literals, so there wouldnt be any need to use escape characters like `\"`. The other difference is it is always displayed as a single line (multiple lines would be reduced down to a single line with a space instead of the new line). Whereas the `"` is opposite and you can use `\n` for new lines if you want to display it as multiple lines

desmond-hui:
Rebase is all weird replacing it with this [PR](https://github.com/FlipsideCrypto/solana-models/pull/682) instead

PR #683, opened 2024-10-30
Hot fix/gold validator tests

tarikceric:
Remove `not null` tests for columns where the API returns nulls
image


desmond-hui (commented):
Can we change these to thresholds instead of completely removing them? I assume a large majority of the records would have these fields populated?

tarikceric:
> Can we change these to thresholds instead of completely removing them? I assume a large majority of the records would have these fields populated?

great point, updated w threshold tests

desmond-hui (approved):

PR #685, opened 2024-10-31
Migrate get solscan blocks to 2.0

desmond-hui:
- Add streamline requests for solscan blocks
  - Sends 10k at a time to streamline, with 1 worker requesting 1k blocks per run. This is similar to the existing workflow where we grab 1k per dbt run command. We are limited by solscan rate limits so can't do more than this.
- Due to API rate limits, deprecate old model in the same step as turning on the new pipeline
  - I will disable the existing workflow manually after merge in case we need to rollback. I will send a separate PR to remove it

```
00:27:57  Running macro `if_data_call_function`: Calling udf streamline.udf_bulk_rest_api_v2 with params: 
{
  "external_table": "solscan_blocks_2",
  "producer_batch_size": "1000",
  "sql_limit": "10000",
  "worker_batch_size": "1000"
}
00:28:07  1 of 1 OK created sql view model streamline.solscan_blocks ..................... [SUCCESS 1 in 10.13s]
```

Silver model updated on M
`dbt run -s silver___blocks_tx_count -t dev`
```
16:27:21  1 of 1 OK created sql incremental model silver._blocks_tx_count ................ [SUCCESS 1852 in 6.51s]
```

Data in DEV
```
select *
from solana_dev.bronze.streamline_solscan_blocks_2;

select *
from solana_dev.silver._blocks_tx_count
where _inserted_timestamp > '2024-10-31 14:00:00';
```


models/streamline/core/realtime/streamline__solscan_blocks.sql
```diff
@@ -0,0 +1,53 @@
+{{ config(
+    materialized = 'view',
+    post_hook = fsc_utils.if_data_call_function_v2(
+        func = 'streamline.udf_bulk_rest_api_v2',
+        target = "{{this.schema}}.{{this.identifier}}",
+        params ={ "external_table" :"solscan_blocks_2",
+        "sql_limit" :"10000",
+        "producer_batch_size" :"1000",
+        "worker_batch_size" :"1000",
+        "sql_source" :"{{this.identifier}}" }
```
eric-laurello, line 10: do you care about the order by here?

desmond-hui: The view will naturally keep the DESC order because of the QUALIFY
```
QUALIFY 
        row_number() OVER (ORDER BY b.block_id DESC) <= {{ producer_limit_size }}
```

```
WITH base AS (
    SELECT 
        b.block_id
    FROM 
        solana.silver.blocks b
    LEFT OUTER JOIN 
        solana.silver._blocks_tx_count b2
        ON b.block_id = b2.block_id
    WHERE 
        b.block_id >= 226000000
        AND b2.block_id IS NULL
        AND b.block_timestamp::DATE <= current_date
),
block_ids AS (
    SELECT 
        block_id 
    FROM 
        base b
    QUALIFY 
        row_number() OVER (ORDER BY b.block_id DESC) <= 10000
)
SELECT
    block_id,
FROM
    block_ids
```

eric-laurello (approved):

PR #686, opened 2024-10-31
model and workflow

tarikceric:
- Model to store updated consume_units for transactions
- set to run 2x every 5 mins based on incrementals below

```
04:19:34  Finished running 1 incremental model, 11 project hooks in 0 hours 2 minutes and 25.83 seconds (145.83s).
04:25:00  Finished running 1 incremental model, 11 project hooks in 0 hours 2 minutes and 21.69 seconds (141.69s).

```


models/silver/backfill/silver__transactions_units_consumed.sql
desmond-hui, line 1: Can we put `backfill` prefix in the model name

models/silver/backfill/silver__transactions_units_consumed.sql
```diff
@@ -0,0 +1,53 @@
...
+    tx_id,
+    compute_units_consumed as units_consumed,
+    _partition_id,
+    _inserted_timestamp,
+    {{ dbt_utils.generate_surrogate_key(['tx_id']) }} AS transactions_id,
+    SYSDATE() AS inserted_timestamp,
+    SYSDATE() AS modified_timestamp,
+    '{{ invocation_id }}' AS _invocation_id
+FROM
+    pre_final
```
desmond-hui, line 53: This is fine, but any reason this isnt just one query w/o the CTE?

models/silver/backfill/silver__transactions_units_consumed.sql
```diff
@@ -0,0 +1,53 @@
...
+        materialized="incremental",
+        unique_key=['tx_id'],
+        tags=['units_consumed_backfill']
+    )
+}}
+
+{% if execute %}
+    {% set next_partition_query %}
+    {% if is_incremental() %}
+        SELECT max(_partition_id)+1, max(_partition_id)+5 FROM {{ this }}
```
desmond-hui, line 12: We should set a global max partition since we are using the value from the node already?

models/silver/backfill/silver__transactions_units_consumed.sql
```diff
@@ -0,0 +1,53 @@
...
+    {% else %}
+        SELECT 24239, 24239 /* When computeUnitsConsumed first appears in node response */
```
desmond-hui, line 14: This goes pretty far back so I would just be on the lookout for large partitions (think times we had to backfill data due to outages or whatever) where this could be an issue.

Screenshot 2024-10-31 at 8 43 30 AM

I am running this just for partition `103410` and it is taking over 10mins already. I think it's worth it to test it out on a larger machine and find some sort of cutoff in terms of # of records in a partition and handle those separately

desmond-hui (changes requested):

models/silver/backfill/silver__transactions_units_consumed.sql
tarikceric: renamed

models/silver/backfill/silver__transactions_units_consumed.sql
tarikceric: simplified it

models/silver/backfill/silver__transactions_units_consumed.sql
tarikceric: added

models/silver/backfill/silver__transactions_units_consumed.sql
tarikceric: added list of partitions to exclude and run manually after

desmond-hui (approved):

PR #691, opened 2024-11-01
Migrate validator list to streamline 2.0

desmond-hui:
- Migrate validator list requests to streamline 2.0
  - Data is unflattened. Streamline 2.0 doesnt seem to be able to flatten when the response is an array w/ no property name. Data is small though, shouldnt be any issues flattening in SF
- Fix workflow naming for streamline validator vote program accounts

`dbt run --vars '{STREAMLINE_INVOKE_STREAMS: True}' -s streamline__validators_list_2`
```
16:54:38  Running macro `if_data_call_function`: Calling udf streamline.udf_bulk_rest_api_v2 with params: 
{
  "external_table": "validators_list_2",
  "producer_batch_size": "1",
  "sql_limit": "1",
  "worker_batch_size": "1"
}
16:54:46  1 of 1 OK created sql view model streamline.validators_list_2 .................. [SUCCESS 1 in 8.48s]
```

Data in DEV
```
select *
from solana_dev.bronze.streamline_validators_list_2;
```

### DIFF
--- a/.github/workflows/dbt_run_streamline_validators_list_snapshot.yml
+++ b/.github/workflows/dbt_run_streamline_validators_list_snapshot.yml
@@ -1,0 +1,46 @@
+name: dbt_run_streamline_validators_list_snapshot
+run-name: dbt_run_streamline_validators_list_snapshot
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Runs 01:22 daily (see https://crontab.guru)
+    - cron: '22 1 * * *'
+    
+env:
+  DBT_PROFILES_DIR: "${{ vars.DBT_PROFILES_DIR }}"
+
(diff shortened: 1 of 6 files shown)


.github/workflows/dbt_run_streamline_validators_list_snapshot.yml
```diff
@@ -0,0 +1,46 @@
+name: dbt_run_streamline_validators_list_snapshot
+run-name: dbt_run_streamline_validators_list_snapshot
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Runs 01:22 daily (see https://crontab.guru)
+    - cron: '22 1 * * *'
```
eric-laurello, line 8: maybe stagger these slightly? I think all three validator ones kick off at the same minute

eric-laurello (approved):

PR #692, opened 2024-11-04
Migrate validator rank metadata

desmond-hui:
- Add streamline 2.0 pipeline for validator metadata
  - Old pipeline was through solscan, but that API is dead (it has been for a bit) with no replacement as far as I can tell
  - New api is through https://topvalidators.app/ and the only thing "missing" in the data is the `validator_rank`. I found that the ranking is just by the # of active stake in descending order. We can do that downstream to get the ranking value that matches solscan's old API response. You can see the solscan ranking [here](https://solscan.io/validator?page=1)

`dbt run -s streamline__validator_metadata_2 --vars '{"STREAMLINE_INVOKE_STREAMS": True}' -t dev`
```
16:24:24  Running macro `if_data_call_function`: Calling udf streamline.udf_bulk_rest_api_v2 with params: 
{
  "external_table": "validator_metadata_2",
  "producer_batch_size": "1",
  "sql_limit": "1",
  "worker_batch_size": "1"
}
16:24:27  1 of 1 OK created sql view model streamline.validator_metadata_2 ............... [SUCCESS 1 in 3.45s]
```

Data in DEV
```
select
    r.value :identity :: STRING AS node_pubkey,
    r.value :commission :: INTEGER AS commission,
    r.value :vote_identity :: STRING AS vote_pubkey,
    r.value :activated_stake :: FLOAT AS stake,
    row_number() over (order by stake desc) AS validator_rank
from solana_dev.bronze.streamline_validator_metadata_2 AS v,
table(flatten(data::array)) AS r
```

### DIFF
--- a/models/streamline/core/snapshot/streamline__validator_metadata_2.sql
+++ b/models/streamline/core/snapshot/streamline__validator_metadata_2.sql
@@ -1,0 +1,29 @@
+{{ config (
+    materialized = "view",
+    post_hook = fsc_utils.if_data_call_function_v2(
+        func = 'streamline.udf_bulk_rest_api_v2',
+        target = "{{this.schema}}.{{this.identifier}}",
+        params ={ 
+            "external_table" :"validator_metadata_2",
+            "sql_limit" :"1",
+            "producer_batch_size" :"1",
+            "worker_batch_size" :"1",
+            "sql_source" :"{{this.identifier}}",
+        }
+    )
+) }}
+
+SELECT
+    replace(current_date::string,'-','_') AS partition_key, -- Issue with streamline handling `-` in partition key so changing to `_`
+    '{{ invocation_id }}' AS invocation_id,
+    {{ target.database }}.live.udf_api(
+        'GET',
+        '{Service}/validators',
+        OBJECT_CONSTRUCT(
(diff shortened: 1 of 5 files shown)


models/streamline/core/snapshot/streamline__validator_metadata_2.sql
```diff
@@ -0,0 +1,29 @@
...
+            "sql_limit" :"1",
+            "producer_batch_size" :"1",
+            "worker_batch_size" :"1",
+            "sql_source" :"{{this.identifier}}",
+        }
+    )
+) }}
+
+SELECT
+    replace(current_date::string,'-','_') AS partition_key, -- Issue with streamline handling `-` in partition key so changing to `_`
```
eric-laurello, line 17: going to steal this, ran into the same issue on Friday

eric-laurello (approved):

models/streamline/core/snapshot/streamline__validator_metadata_2.sql
desmond-hui: Yep, it's easy to bring it back to `YYYY-MM-DD` date format in external/bronze views

PR #702, opened 2024-11-07
increase batch size; run interval

tarikceric:
Updates to backfill fast as currently it would take a few weeks just for the 1 month of data we need
- increase batch size to match real time limit
- Schedule more runs per hour -- single run has been taking ~5 mins

Test run in dev:
- ran for ~4mins
- results from run: Returns 9000 
```
select count(distinct(tx_id)) from solana_dev.silver.nft_compressed_mints
where modified_timestamp = '2024-11-08 16:37:54.837';
```


desmond-hui (commented):
Have these limits been tested in DEV?

desmond-hui (approved):

PR #725, opened 2024-12-04
temporary job to check if all null mints are wsol

desmond-hui:
Run this backfill backwards to ensure all the missing mints are `WSOL` so that we can run a simple update statement. Scheduled workflows have to be in main branch so we are forced to merge this into main even though this table will only live in dev

### DIFF
--- a/models/silver/backfill/silver__transfers_null_mints_fix.sql
+++ b/models/silver/backfill/silver__transfers_null_mints_fix.sql
@@ -1,0 +1,30 @@
+{{
+    config(
+        materialized = 'incremental',
+        full_refresh = false
+    )
+}}
+
+{% if execute and is_incremental() %}
+    {% set next_block_date_to_process_query %}
+        select min(block_timestamp::date)-1 from {{ this }}
+    {% endset %}
+    {% set next_block_date_to_process = run_query(next_block_date_to_process_query)[0][0] %}
+{% endif %}
+
+select 
+    t.block_timestamp,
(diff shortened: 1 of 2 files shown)


models/silver/backfill/silver__transfers_null_mints_fix.sql
```diff
@@ -0,0 +1,30 @@
...
+{% endif %}
+
+select 
+    t.block_timestamp,
+    tr.*,
+    (
+        silver.udf_get_account_balances_index(dest_token_account, t.account_keys) IS NOT NULL
+        OR silver.udf_get_account_balances_index(source_token_account, t.account_keys) IS NOT NULL
+    ) AS is_wsol
+from solana_dev.silver.transfers_null_mints AS tr
```
tarikceric, line 22: does this table contain every null mint in transfers table or is it up to a certain date?

tarikceric (approved):

models/silver/backfill/silver__transfers_null_mints_fix.sql
desmond-hui: Good question! I created it with this, so just after 2023-01-01

```
-- create or replace transient table solana_dev.silver.transfers_null_mints as
-- select tx_id, source_token_account, dest_token_account
-- from solana.silver.transfers
-- where block_timestamp::date >= '2023-01-01'
-- and succeeded
-- and mint is null;
```

PR #726, opened 2024-12-04
Deprecate legacy decode logs pipeline

desmond-hui:
> [!IMPORTANT]  
> This will not be merged until **Monday 12/9 630am PT**

- Remove calls to legacy pipeline
- Modify realtime query to match existing but using `streamline.complete_decoded_logs_2`
- Modify `silver.decoded_logs` to use streamline 2.0 upstream data after cutoff data `2024-12-09`
  - Temporarily extend lookback window
- Modify backfill macro to call streamline 2.0 endpoint

Streamline models run on DEV
`dbt run -s streamline__decode_logs_2_realtime -t dev`
```
16:46:35  1 of 1 OK created sql table model streamline.decode_logs_2_realtime ............ [SUCCESS 1 in 105.39s]
```

Silver model on DEV
`dbt run -s silver__decoded_logs -t dev-2xl`
```
17:07:53  1 of 1 OK created sql incremental model silver.decoded_logs .................... [SUCCESS 403275 in 76.44s]
```

### DIFF
--- a/models/silver/parser/silver__decoded_logs.sql
+++ b/models/silver/parser/silver__decoded_logs.sql
@@ -21,14 +21,15 @@
 
 {% set CUTOVER_DATETIME = modules.datetime.datetime.strptime("2024-07-16 17:00:00", "%Y-%m-%d %H:%M:%S") %}
 {% set use_legacy_logic = False %}
-{% set streamline_2_cutover_datetime = modules.datetime.datetime.strptime("2024-12-04 00:00:00+00:00", "%Y-%m-%d %H:%M:%S%z") %}
+{% set streamline_2_cutover_datetime = modules.datetime.datetime.strptime("2024-12-09 00:00:00+00:00", "%Y-%m-%d %H:%M:%S%z") %}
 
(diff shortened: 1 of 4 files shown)


models/silver/parser/silver__decoded_logs.sql
```diff
@@ -45,12 +50,12 @@
                 log_index,
                 program_id,
                 data,
-                _inserted_timestamp,
+                _inserted_timestamp
             FROM
                 {% if is_incremental() %}
-                {{ ref('bronze__streamline_decoded_logs') }} A
+                    {{ ref('bronze__streamline_decoded_logs') }} A
```
tarikceric, line 56: should these reference `decoded_logs_2` too?

desmond-hui: No, this section of the code is only for some other legacy logic (see the if block that it is a part of) that I don't recall why it exists but streamline 2.0 data would never come before the `CUTOVER_DATETIME` which is `2024-07-16 17:00:00`

PR #738, opened 2024-12-13
add tx_index for data sourced from streamline 2.0

desmond-hui:
- Add `tx_index` column to `silver.transactions`
  - Run DML before merging so that references to `{{ this }}` that selects all the columns in the model will work correctly
  ```
  ALTER TABLE solana.silver.transactions
  ADD COLUMN tx_index INTEGER;
  ```
- Add better filtering on data tests 

`dbt run -s silver__transactions -t dev`
```
00:32:32  1 of 13 OK created sql incremental model silver.transactions ................... [SUCCESS 1507135 in 245.06s]
```

Data in DEV
```
-- No new txs without tx_index
select count(*)
from solana_dev.silver.transactions
where _partition_id > 150217
and tx_index is null;

-- tx index is populated
select block_id, tx_id, tx_index
from solana_dev.silver.transactions
where _partition_id > 150217
and tx_index is not null
limit 10;
```

### DIFF
--- a/models/descriptions/tx_index.md
+++ b/models/descriptions/tx_index.md
@@ -1,0 +1,5 @@
+{% docs tx_index %}
+
+The index of the transaction in the block. Index of 0 is the first transaction executed in the block.
+
+{% enddocs %}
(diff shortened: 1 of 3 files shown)


tarikceric:
update core in another pr?

tarikceric (approved):

PR #741, opened 2024-12-17
Deprecate/orca init pools

tarikceric:
deprecate initialization_pools_orca as logic needs a complete revamp and no new records are coming through
- use view table as reference for dependant tables (including already deprecated models which isn't completely necessary)

incremental runs (excluding deprecated models)

```
19:55:17  1 of 4 OK created sql view model silver.initialization_pools_orca_view ......... [SUCCESS 1 in 2.34s]
19:55:49  2 of 4 OK created sql incremental model silver.liquidity_pool_events_orca ...... [SUCCESS 15803 in 31.91s]
19:55:58  3 of 4 OK created sql incremental model silver.mints_orca_whirlpool ............ [SUCCESS 1 in 8.37s]
19:57:19  4 of 4 OK created sql incremental model silver.pool_transfers_orca_whirlpool ... [SUCCESS 15211 in 89.46s]
```

tests passed
```
20:05:02  Finished running 40 data tests, 7 project hooks in 0 hours 1 minutes and 16.84 seconds (76.84s).
20:05:02  
20:05:02  Completed successfully
20:05:02  
20:05:02  Done. PASS=40 WARN=0 ERROR=0 SKIP=0 TOTAL=40
```

### DIFF
--- a/models/gold/defi/defi__fact_liquidity_pool_actions.yml
+++ b/models/gold/defi/defi__fact_liquidity_pool_actions.yml
@@ -1,7 +1,7 @@
 version: 2
 models:
   - name: defi__fact_liquidity_pool_actions
-    description: Table containing liquidity pools actions on Orca, Raydium, Saber and Meteora. 
+    description: Table containing liquidity pools actions on Orca, Raydium, Saber and Meteora. NOTE - actions for Orca pools created after 11/22/2024 are not tracked in this table, and we are working on updating our Orca data pipeline to collect these.
     recent_date_filter: &recent_date_filter
       config:
         where: modified_timestamp >= current_date - 7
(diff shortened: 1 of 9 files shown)


desmond-hui (commented):
How is this affecting models that are surfaced to the public?

tarikceric:
> How is this affecting models that are surfaced to the public?

These LP models feed into `defi.fact_liquidity_pool_actions`, and the deprecated initialization_pools_orca hasn't been returning records in about a month (besides two incorrect records from last week that should be deleted). The existing pools in the deprecated model are still referenced in other models, so lp actions for those pools should still be captured downstream, but newly created pools wouldn't be --and this older logic doesn't seem like it'd be capturing those pool initializations anyway

desmond-hui:
So basically the orca related actions in `defi.fact_liquidity_pool_actions` are likely to remain stale until we find a fix right? Is there any way to surface that information to the consumers?

desmond-hui (approved):

PR #749, opened 2025-01-07
remove cnft backfill

tarikceric:
DBT commands:

1. seed workflow file: dbt seed -s github_actions__workflows -t prod
2. rebuild views: dbt run -m models/github_actions --full-refresh -t prod
3. rebuild and start asks: dbt run-operation fsc_utils.create_gha_tasks --vars '{"START_GHA_TASKS":True}' -t prod

To drop task:
drop task solana.github_actions.TRIGGER_DBT_RUN_NFT_COMPRESSED_MINTS_BACKFILL;


desmond-hui (approved):

desmond-hui:
Do we want to keep the actual workflow file around? If so, is there a way for us to mark that it is `deprecated` or `paused`?

tarikceric:
> Do we want to keep the actual workflow file around? If so, is there a way for us to mark that it is `deprecated` or `paused`?

We should keep it around in case we need to do backfills at some other point. We can add a comment at the top of the yaml to note that its deprecated, and some other options I've found are:
- rename the file to `,yml.disabled`
- comment out the `on` trigger

I like just adding a comment but wdyt? Ill ask the team channel if theyve done something similar

PR #760, opened 2025-01-09
An 5651/marinade ez swaps

tarikceric:
table only containing MNDE and MSOL token swaps

- fr on 2xl:
`20:55:26  Finished running 1 incremental model, 7 project hooks in 0 hours 1 minutes and 27.96 seconds (87.96s).`

- test
`21:05:49  Finished running 28 data tests, 7 project hooks in 0 hours 0 minutes and 59.93 seconds (59.93s).`

- inc run on medium
`21:08:45  1 of 1 OK created sql incremental model marinade.ez_swaps ...................... [SUCCESS 1116 in 12.37s]`

### DIFF
--- a/models/gold/marinade/marinade__ez_swaps.sql
+++ b/models/gold/marinade/marinade__ez_swaps.sql
@@ -1,0 +1,147 @@
+{{ config(
+    materialized = 'incremental',
+    meta = { 'database_tags': { 'table': { 'PURPOSE': 'SWAPS' }}},
+    unique_key = ['marinade_ez_swaps_id'],
+    incremental_predicates = ["dynamic_range_predicate", "block_timestamp::date"],
+    cluster_by = ['block_timestamp::DATE', 'program_id'],
+    merge_exclude_columns = ["inserted_timestamp"],
+    post_hook = enable_search_optimization('{{this.schema}}', '{{this.identifier}}', 'ON EQUALITY(tx_id, swapper, swap_from_mint, swap_to_mint, program_id)'),
+    tags = ['scheduled_non_core']
+) }}
+
+WITH swaps_non_agg AS (
+    SELECT 
+        block_id,
+        block_timestamp,
+        tx_id,
+        swapper,
+        swap_from_mint,
+        swap_from_symbol,
+        swap_from_amount,
+        swap_from_amount_usd,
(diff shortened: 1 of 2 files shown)


models/gold/marinade/marinade__ez_swaps.sql
```diff
@@ -0,0 +1,155 @@
...
+        program_id,
+        swap_program AS platform,
+        ez_swaps_id AS marinade_ez_swaps_id,
+        inserted_timestamp,
+        modified_timestamp
+    FROM 
+        {{ ref('defi__ez_dex_swaps') }}
+    WHERE 
+        (swap_from_mint IN ('MNDEFzGvMt87ueuHvVU9VcTqsAP5b3fTGPsHuuPA5ey', 'mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So') OR swap_to_mint IN ('MNDEFzGvMt87ueuHvVU9VcTqsAP5b3fTGPsHuuPA5ey', 'mSoLzYCxHdYgdzU16g5QSh3i5K3z3KZK7ytfqcJm7So'))
+        AND tx_id NOT IN (SELECT tx_id FROM swaps_jupiter)
```
desmond-hui, line 110: This will only work if the swap is inside of the subset that is being loaded in the incremental. I think the logic needs to be revisited

desmond-hui (changes requested):

models/gold/marinade/marinade__ez_swaps.sql
tarikceric: this is true. Would we use jupiter_inner to identify the ones that are part of a jupiter swap and exclude those

desmond-hui: That sounds like a better approach but I am still concerned w/ some timing issues that would surface jupiter inner swaps in the single dex swaps.

tarikceric: updated logic - attempting to get a 3hour buffer for the modified_timestamp in the `fact_swaps_jupiter_summary`, and then use that block_range filter when selecting from the `marinade_swaps` table so we always get the related swaps

eric-laurello (approved):

PR #772, opened 2025-01-15
Add decoder backfill helper for single dates

desmond-hui:
Add macro helpers for backfilling decoder instructions and logs for a specific `block_timestamp::date`. This can be useful for situations where there are significant raw data ingestion delays as the current real-time streamline requests only have a 2 day lookback.

`dbt run-operation decoded_logs_backfill_retry_single_date_all_programs --args '{"backfill_date": "2024-12-15"}' -t dev`
Screenshot 2025-01-14 at 4 09 46 PM

`dbt run-operation decoded_instructions_backfill_single_date_all_programs --args '{"backfill_date": "2024-12-15"}' -t dev`
Screenshot 2025-01-14 at 4 10 40 PM

### DIFF
--- a/macros/helpers/decoded_instructions_backfill_helpers.sql
+++ b/macros/helpers/decoded_instructions_backfill_helpers.sql
@@ -306,6 +306,7 @@
                     ON c.program_id = b.program_id
                 WHERE 
                     _inserted_timestamp >= '{{ retry_start_timestamp }}'
+                    AND block_id BETWEEN {{ start_block }} AND {{ end_block }}
             )
             SELECT
                 e.program_id,
--- a/macros/helpers/decoded_instructions_backfill_helpers.sql
+++ b/macros/helpers/decoded_instructions_backfill_helpers.sql
@@ -6,7 +6,7 @@
         where program_id = '""" ~ program_id ~ """';""").columns %}
     {% set min_block_id = result_cols[0].values()[0] | int %}
     {% set max_block_id = result_cols[1].values()[0] | int %}
-    {% set step = 10000000 %}
+    {% set step = 2000000 %}
 
     {% for i in range(min_block_id, max_block_id, step) %}
         {% if i == min_block_id %}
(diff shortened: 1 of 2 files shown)


macros/helpers/decoded_instructions_backfill_helpers.sql
```diff
@@ -200,6 +200,136 @@
...
+            retry_events AS (
+                SELECT
+                    e.program_id,
+                    e.tx_id,
+                    e.index,
+                    NULL as inner_index,
+                    e.instruction,
+                    e.block_id,
+                    e.block_timestamp,
+                    {{ dbt_utils.generate_surrogate_key(['e.block_id','e.tx_id','e.index','inner_index','e.program_id']) }} as id
```
tarikceric, line 264: nit: do we need inner_index if its always null?

macros/helpers/decoded_logs_backfill_helpers.sql
```diff
@@ -121,6 +121,145 @@
     {% endfor %}
 {% endmacro %}
 
+{% macro decoded_logs_backfill_retry_single_date_all_programs(backfill_date, priority=None) %}
```
tarikceric, line 124: nit: should it have 'retry' in the name

macros/helpers/decoded_instructions_backfill_helpers.sql
```diff
@@ -200,6 +200,136 @@
...
+                    {{ ref('silver__events_inner') }} AS e
+                JOIN 
+                    idl_in_play b
+                    ON e.program_id = b.program_id
+                WHERE
+                    e.block_timestamp::date = '{{ backfill_date }}'
+                    AND e.succeeded
+                    AND (
+                        (
+                            e.program_id IN ('FLASH6Lo6h3iasJKWDs2F8TkW2UKf3s15C8PMGuVfgBn','SNPRohhBurQwrpwAptw1QYtpFdfEKitr4WSJ125cN1g','GovaE4iu227srtG2s3tZzB4RmWBzw8sTwrCLZz7kN7rY','JUP6LkbZbjS1jKKwapdHNy74zcZ3tLUZoi5QNyVTaV4','DCA265Vj8a9CEuX1eb1LWRnDT7uK6q1xMipnNyatn23M','PERPHjGBqRHArX4DySjwM6UJHiR3sWAatqfdBS2qQJu','LBUZKhRxPF3XUpBCjp4YzTKgLccjZhTSDM9YuVaPwxo','PhoeNiXZ8ByJGLkxNfZRnkUfjvmuYqLR89jjFHGqdXY','6EF8rrecthR5Dkzon8Nwu78hRvfCKubJ14M5uBEwF6P')
```
tarikceric, line 293: whats the criteria for manually adding programs in the future to this?

macros/helpers/decoded_logs_backfill_helpers.sql
desmond-hui: TL;DR - Yes

There are some technical and semantic/non-technical reasons as to why I think we need to include it. First, while true it is always `NULL` for the top level event, it is not true for the inner events which this CTE is unioning to. So for the CTE, `inner_index` is not always `NULL`.

The way `dbt_utils.generate_surrogate_key()` works makes it so the generated key is different even if one of the columns is null. As in...assuming you have two columns, [A,B] and assume that B is always `NULL`. `dbt_utils.generate_surrogate_key('A','B')` gives a different answer than `dbt_utils.generate_surrogate_key('A')`.

Non-technical reason...we should think of the CTE as a table. We would not define a surrogate key differently just because some X records always has some column B as `NULL`, while some Y records always has some column B as `non-NULL`. Because we _need_ to have `inner_index` in the surrogate key generation for some of the rows in the table then we should define it that way for the whole table.

macros/helpers/decoded_instructions_backfill_helpers.sql
desmond-hui: TL;DR - Yes

There are some technical and semantic/non-technical reasons as to why I think we need to include it. First, while true it is always `NULL` for the top level event, it is not true for the inner events which this CTE is unioning to. So for the CTE, `inner_index` is not always `NULL`.

The way `dbt_utils.generate_surrogate_key()` works makes it so the generated key is different even if one of the columns is null. As in...assuming you have two columns, [A,B] and assume that B is always `NULL`. `dbt_utils.generate_surrogate_key('A','B')` gives a different answer than `dbt_utils.generate_surrogate_key('A')`.

Non-technical reason...we should think of the CTE as a table. We would not define a surrogate key differently just because some X records always has some column B as `NULL`, while some Y records always has some column B as `non-NULL`. Because we _need_ to have `inner_index` in the surrogate key generation for some of the rows in the table then we should define it that way for the whole table.

macros/helpers/decoded_instructions_backfill_helpers.sql
desmond-hui: These are all programs that generate log that are output as encoded data within an `inner_instruction`. So if you know of a program "we" care about that does this, it should be included. It doesnt break anything if it isnt included but we end up sending many requests that will fail since those records need to be sent to the `logs decoder`

tarikceric (approved):

PR #775, opened 2025-01-21
An 5383/me decoded update

tarikceric:
Create Magic Eden v2 sales table from decoded instructions. ME supports payment in tokens beside SOL so add column for `currency_address`.
- deprecate original table
- in gold `nft.fact_nft_sales `:
     -  add `currency_address` column, and set as SOL for all other models
     - add new model -- since the unique key for both the legacy and new model is the same, the old records should be updated in an incremental run

To update gold:
1. Update the currency_address to SOL
```
update solana_dev.nft.fact_nft_sales
set currency_address = 'So11111111111111111111111111111111111111111'
where currency_address is null
```

Runs:

- Incremental on medium for silver and gold:
`21:26:57  1 of 1 OK created sql incremental model silver.nft_sales_magic_eden_v2_decoded . [SUCCESS 170 in 15.55s]`
`21:32:44  1 of 1 OK created sql incremental model nft.fact_nft_sales ..................... [SUCCESS 8053500 in 28.10s]`

- FR on 2xl for `magic_eden_v2_decoded`:
`16:35:36  Finished running 1 incremental model, 7 project hooks in 0 hours 7 minutes and 21.09 seconds (441.09s).
`

- Tests for silver and gold:
```
16:53:45  Finished running 53 data tests, 7 project hooks in 0 hours 1 minutes and 22.95 seconds (82.95s).
16:53:46  
16:53:46  Completed successfully
16:53:46  
16:53:46  Done. PASS=53 WARN=0 ERROR=0 SKIP=0 TOTAL=53
```


models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.sql
```diff
@@ -0,0 +1,151 @@
...
+{% if is_incremental() %}
+AND _inserted_timestamp >= (
+    SELECT 
+        MAX(_inserted_timestamp) - INTERVAL '1 hour'
+    FROM 
+        {{ this }}
+)
+{% else %}
+        -- AND block_timestamp::DATE > '2024-10-01' -- temp test but update to where decoded instructions are good
+        and _inserted_timestamp >= '2023-11-15'
```
desmond-hui, line 43: More of a question...Is there a block_timestamp cutoff for when we should be using decoded data vs. the legacy data?

models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.yml
```diff
@@ -0,0 +1,77 @@
+version: 2
+models:
+  - name: silver__nft_sales_magic_eden_v2_decoded
+    description: intermediary model for sales on Magic Eden V2 using decoded instructions
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - TX_ID
```
desmond-hui, line 8: nit: can be `unique` test for the column `tx_id`

models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.yml
```diff
@@ -0,0 +1,77 @@
...
+          - not_null: *recent_date_filter
+      - name: TX_ID
+        description: "{{ doc('tx_id') }}"
+        tests:
+          - not_null: *recent_date_filter
+      - name: SUCCEEDED
+        description: "{{ doc('tx_succeeded') }}"
+        tests:
+          - not_null: *recent_date_filter
+      - name: PROGRAM_ID
```
desmond-hui, line 32: Missing `index` and `inner_index`

desmond-hui (changes requested):
Do we need an `ALTER TABLE` statement for gold?

models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.yml
tarikceric: removed this and added `unique` test on `tx_id` -- should we keep the `unique` test on `nft_sales_magic_eden_v2_decoded_id`?

models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.sql
tarikceric: I initially did have `2023-01-09` as the cutoff to use the new data -- but there's a handful of dates where there's a few nfts still missing, and i think its just decoding issues. Its only 45 in total over the next two years.

with a as (
select * from solana_dev.silver.nft_sales_magic_eden_v2
where tx_id not in (select tx_id from solana_dev.silver.nft_sales_magic_eden_v2_decoded)
and succeeded
)
select count(*), block_timestamp::date
from a
group by 2;

Those handful that are missing are the reason i didn't do a cutoff, because the unique id would've updated the ones we do have while still keep the ones from the legacy model. Also why i didn't plan on using a FR.

models/gold/nft/nft__fact_nft_sales.sql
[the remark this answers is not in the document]
tarikceric: Explained the situation a bit in the next comments and would like your thoughts

models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.yml
desmond-hui: Yep, keep both

models/silver/nfts/silver__nft_sales_magic_eden_v2_decoded.sql
desmond-hui: > Those handful that are missing are the reason i didn't do a cutoff, because the unique id would've updated the ones we do have while still keeping the ones from the legacy model.

If they arent in the new table, wouldn't we still use the old data?

> The new table is the most complete so i'd want to use that data as much as possible, with the earliest cutoff date being 2023-01-09 , but how do you suggest I make sure those few missing NFT sales from the legacy data are still added in an FR?

Can we do a `FULL OUTER JOIN` on both models then select the columns using `IFF(new.tx_id IS NULL, use old data, use new data) AS my_column`?

tarikceric: > If they arent in the new table, wouldn't we still use the old data?

I was thinking about if we do an FR. You're right that just continuing with incremental would ensure we use the old data, but that's the issue with cutoff. I want to use `2023-01-09` since that's when most of the decoding is correct. But theres a handful of missing NFTs up to `2024-03-15` and i don't want to set the cutoff that far forward, since the new table is getting sales not in the old table

> Can we do a FULL OUTER JOIN  on both models then select the columns....

Unsure what you mean by selecting the columns? also are you thinking to use the full outer join in the incremental portion of the nft.fact_nft_sales logic, or just have it once in the non-incremental partition?

tarikceric: Updated _view to be a combination of new and legacy decoded, up to` 2024-03-16`. This is the date from where we only want to use the new data, which is reflected in the fact table.

models/gold/nft/nft__fact_nft_sales.sql
```diff
@@ -594,3 +607,32 @@ FROM
...
+    inserted_timestamp,
+    modified_timestamp,
+FROM
+    {{ ref('silver__nft_sales_magic_eden_v2_decoded') }}
+{% if is_incremental() %}
+WHERE
+    modified_timestamp >= '{{ max_modified_timestamp }}'
+{% else %}
+WHERE
+    block_timestamp::date >= '2024-03-16'
```
desmond-hui, line 637: nit: Change to a jinja var and reference it here and in the other location

models/silver/nfts/silver__nft_sales_magic_eden_v2_view.sql
```diff
@@ -0,0 +1,35 @@
...
+    ON
+        d.tx_id = v.tx_id
+)
+
+SELECT
+    *
+FROM
+    base
+WHERE
+    block_timestamp::DATE < '2024-03-16'
```
desmond-hui, line 34: nit: use jinja var and reference that here

desmond-hui (approved):

PR #782, opened 2025-01-29
update mayan bridge

tarikceric:
Create new silver mayan bridge table using decoded data. No events using the old program_id have landed in the table since `2024-12-23`, and the decoded table allows us to get the bridge events using the newer structure.
- convert old table into view
- Legacy and new table do not overlap so directly added it into the gold view

Runs:
- FR on 2xl
`00:12:26  1 of 1 OK created sql incremental model silver.bridge_mayan_transfers_decoded .. [SUCCESS 1 in 195.91s]
`
- incremental on med
`00:31:50  1 of 1 OK created sql incremental model silver.bridge_mayan_transfers_decoded .. [SUCCESS 1426 in 31.48s]
`
- test:
```
00:15:32  Finished running 16 data tests, 7 project hooks in 0 hours 0 minutes and 49.62 seconds (49.62s).
00:15:32  
00:15:32  Completed successfully
00:15:32  
00:15:32  Done. PASS=16 WARN=0 ERROR=0 SKIP=0 TOTAL=16
```

### DIFF
--- a/models/silver/bridges/silver__bridge_mayan_transfers_decoded.sql
+++ b/models/silver/bridges/silver__bridge_mayan_transfers_decoded.sql
@@ -1,0 +1,185 @@
+-- depends_on: {{ ref('silver__decoded_instructions_combined') }}
+
+{{ config(
+    materialized = 'incremental',
+    incremental_predicates = ["dynamic_range_predicate", "block_timestamp::date"],
+    merge_exclude_columns = ["inserted_timestamp"],
+    unique_key = "bridge_mayan_transfers_decoded_id",
+    cluster_by = ['block_timestamp::DATE', 'modified_timestamp::DATE'],
+    post_hook = enable_search_optimization('{{this.schema}}','{{this.identifier}}','ON EQUALITY(tx_id,user_address,mint)'),
+    tags = ['scheduled_non_core']
+) }}
+
+{% if execute %}
+    {% set base_query %}
+    CREATE OR REPLACE TEMPORARY TABLE silver.bridge_mayan_transfers_decoded__intermediate_tmp AS
+    SELECT
+        block_timestamp,
+        block_id,
+        tx_id,
+        succeeded,
+        signers,
+        index,
+        inner_index,
+        program_id,
+        event_type,
+        decoded_instruction,
+        _inserted_timestamp
+    FROM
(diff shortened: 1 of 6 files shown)


desmond-hui:
> Legacy and new table do not overlap so directly added it into the gold view

Why don't they overlap? The `decoded` table starts at `2024-06-21 10:35:38.000` and the legacy table ends at `2024-12-23 08:13:08.000`

tarikceric:
> > Legacy and new table do not overlap so directly added it into the gold view
> 
> Why don't they overlap? The `decoded` table starts at `2024-06-21 10:35:38.000` and the legacy table ends at `2024-12-23 08:13:08.000`

I meant the tx's don't overlap -- the bridge events picked up by the legacy table arent picked up by the new one, and vice versa. They use different program_ids and mayan used both types of events before they completely stopped using the legacy on on `2024-12-23'.

models/silver/bridges/silver__bridge_mayan_transfers_decoded.sql
```diff
@@ -0,0 +1,185 @@
...
+        b.mint,
+        a._inserted_timestamp
+    FROM
+        bridge_to a
+    LEFT JOIN
+        transfers b
+    ON
+        a.tx_id = b.tx_id
+        AND a.trader_dest = b.tx_to
+        AND (a.mint_to = b.mint OR a.mint_to = 'So11111111111111111111111111111111111111112')
```
desmond-hui, line 163: Just curious why we need the `OR a.mint_to = 'So11111111111111111111111111111111111111112'` here. Would't the `WSOL` amount be in transfers?

desmond-hui (commented):
Just a couple of questions

models/silver/bridges/silver__bridge_mayan_transfers_decoded.sql
tarikceric: In Mayan the mint_to in the decoded instructions is `So11111111111111111111111111111111111111112`, but in the event_type I'm using they only show the transfer of native SOL, which from my interpretation gets converted to wSOL, and is also consistent with only wSOL existing in the legacy table as well.

desmond-hui (approved):

PR #785, opened 2025-01-29
stage block txs extract columns from value, don't include it in output

desmond-hui:
- Exclude the `value` column from the output for `bronze.stage_block_txs_2`
  - Extract only the data we need downstream from the value column 
  - Reduces the amount written by about half which also then decreases run time

Existing
Screenshot 2025-01-29 at 9 29 08 AM
New
Screenshot 2025-01-29 at 9 29 16 AM

Existing
```
17:28:23  1 of 1 OK created sql incremental model bronze.stage_block_txs_2 ............... [SUCCESS 2100383 in 155.33s]
```
New
```
17:27:53  1 of 1 OK created sql incremental model bronze.stage_block_txs_2 ............... [SUCCESS 2100383 in 131.10s]
```

### DIFF
--- a/models/bronze/stage/bronze__stage_block_txs_2.sql
+++ b/models/bronze/stage/bronze__stage_block_txs_2.sql
@@ -8,9 +8,14 @@
 }}
 
 SELECT
-    *
+    to_timestamp_ntz(t.value:"result.blockTime"::int) AS block_timestamp,
+    t.block_id,
+    t.value:array_index::int AS tx_index,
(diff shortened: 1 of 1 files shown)


tarikceric:
would adding any logic for skipped blocks be useful?

tarikceric (approved):

desmond-hui:
> would adding any logic for skipped blocks be useful?

It's built in to bronze already

PR #790, opened 2025-02-04
use timestamp filter

tarikceric:
For models where we need to reselect all tx_id's in the the incremental subset, include the block_timestamp in the initial select and use that as a filter to scan less of the decoded table
- this step seems to take the most time and which is in a number of models

The temp table creation runs considerable faster with the additional logic - example runs below:

- new logic: https://app.snowflake.com/us-east-1/vna27887/#/compute/history/queries/01ba2c71-0412-db47-3d4f-8302e81379cf/detail
- existing: https://app.snowflake.com/zsniary/exa10207/#/compute/history/queries/01ba2c60-0412-ddf6-3d4f-8302e811879b/detail

- incremental runs on medium
00:09:36  1 of 6 OK created sql incremental model silver.swaps_inner_intermediate_jupiterv4  [SUCCESS 89 in 42.49s]
00:09:46  4 of 6 OK created sql incremental model silver.swaps_intermediate_raydium_v4_amm  [SUCCESS 949333 in 52.31s]
00:09:47  3 of 6 OK created sql incremental model silver.swaps_intermediate_meteora ...... [SUCCESS 100556 in 53.52s]
00:09:49  5 of 6 OK created sql incremental model silver.swaps_intermediate_jupiterv4_2 .. [SUCCESS 36 in 12.89s]
00:09:56  2 of 6 OK created sql incremental model silver.swaps_inner_intermediate_jupiterv6  [SUCCESS 356574 in 62.25s]
00:10:19  6 of 6 OK created sql incremental model silver.swaps_intermediate_jupiterv6_2 .. [SUCCESS 336684 in 22.61s]

### DIFF
--- a/models/silver/swaps/raydium/silver__swaps_intermediate_raydium_v4_amm.sql
+++ b/models/silver/swaps/raydium/silver__swaps_intermediate_raydium_v4_amm.sql
@@ -13,7 +13,8 @@
         CREATE OR REPLACE TEMPORARY TABLE silver.swaps_intermediate_raydium_v4_amm__intermediate_tmp AS
         WITH distinct_entities AS (
             SELECT DISTINCT
-                tx_id
+                tx_id,
+                block_timestamp
             FROM 
                 {{ ref('silver__decoded_instructions_combined') }}
             WHERE
@@ -33,6 +34,7 @@
                 {% else %}
                     AND _inserted_timestamp :: DATE >= '2024-05-14'
                 {% endif %}
+            GROUP BY 1,2
         )
         /* need to re-select all decoded instructions from all tx_ids in incremental subset 
         in order for the window function to output accurate values */
@@ -59,6 +61,7 @@
                 'swapBaseIn',
                 'swapBaseOut'
             )
+            and d.block_timestamp >= (select min(block_timestamp) from distinct_entities)
     {% endset %}
     {% do run_query(base_query) %}
     {% set between_stmts = fsc_utils.dynamic_range_predicate("silver.swaps_intermediate_raydium_v4_amm__intermediate_tmp","block_timestamp::date") %}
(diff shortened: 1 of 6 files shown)


models/silver/swaps/raydium/silver__swaps_intermediate_raydium_v4_amm.sql
```diff
@@ -33,6 +34,7 @@
                 {% else %}
                     AND _inserted_timestamp :: DATE >= '2024-05-14'
                 {% endif %}
+            GROUP BY 1,2
```
desmond-hui, line 37: Don't need this group by, it's already `SELECT DISTINCT`

models/silver/swaps/raydium/silver__swaps_intermediate_raydium_v4_amm.sql
```diff
@@ -59,6 +61,7 @@
                 'swapBaseIn',
                 'swapBaseOut'
             )
+            and d.block_timestamp >= (select min(block_timestamp) from distinct_entities)
```
desmond-hui, line 64: nit: match formatting in rest of the file

desmond-hui (approved):
Nice find! This makes sense and does seem to show great deal of run time improvements.

Just a note on the samples you linked. I would try to make sure it is set up to compare apples to apples as much as possible. The existing query is outputting different number of txs than the new one. In this case I don't think it has much effect on the conclusion.

models/silver/swaps/raydium/silver__swaps_intermediate_raydium_v4_amm.sql
tarikceric: updated

models/silver/swaps/raydium/silver__swaps_intermediate_raydium_v4_amm.sql
tarikceric: updated

desmond-hui (approved):

PR #801, opened 2025-02-11
filter with block id buffer

tarikceric:
Limit the scan of `blocks` and `backfill_transactions_index` by filtering the majority of blocks already processed via the min block_id which should be the last processed block
- added a buffer of 100k blocks in case 

- dbt run --vars '{"STREAMLINE_INVOKE_STREAMS": True}' -s streamline__block_txs_index -t dev
`22:01:29  1 of 1 OK created sql view model streamline.block_txs_index .................... [SUCCESS 1 in 7.05s]`
compared to last prod run
`17:32:38  2 of 2 OK created sql view model streamline.block_txs_index .................... [SUCCESS 1 in 256.53s]
`

### DIFF
--- a/models/streamline/core/realtime/streamline__block_txs_index.sql
+++ b/models/streamline/core/realtime/streamline__block_txs_index.sql
@@ -15,19 +15,27 @@
     )
 ) }}
 
+{% if execute %}
+    {% set min_block_id_buffer_query %}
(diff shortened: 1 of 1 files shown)


desmond-hui:
> theoretically all the blocks after that should be complete so there shouldnt be a need to scan through those

I like the idea but in this instance that assumption does not hold true

PR #803, opened 2025-02-13
add program to cluster

tarikceric:
add `program_id` and `instruction_program_id` to `events_inner` clustering

### DIFF
--- a/models/silver/core/silver__events_inner.sql
+++ b/models/silver/core/silver__events_inner.sql
@@ -2,7 +2,7 @@
     materialized = 'incremental',
     unique_key = ['block_id', 'tx_id', 'instruction_index', 'inner_index'],
     incremental_predicates = ["dynamic_range_predicate", "block_timestamp::date"],
-    cluster_by = ['block_timestamp::DATE','modified_timestamp::DATE','ROUND(block_id, -3)','program_id','instruction_program_id'],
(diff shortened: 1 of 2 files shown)


models/silver/core/silver__events_inner.sql
```diff
@@ -2,7 +2,7 @@
     materialized = 'incremental',
     unique_key = ['block_id', 'tx_id', 'instruction_index', 'inner_index'],
     incremental_predicates = ["dynamic_range_predicate", "block_timestamp::date"],
-    cluster_by = ['block_timestamp::DATE','modified_timestamp::DATE','ROUND(block_id, -3)'],
+    cluster_by = ['block_timestamp::DATE','modified_timestamp::DATE','ROUND(block_id, -3)','program_id','instruction_program_id'],
```
eric-laurello, line 5: any reason to add to silver?

tarikceric: ilver tables that reference `events_inner` use those columns so figured it might speed those incrementals up, but I'm fine with not adding it

eric-laurello: yeah unless you have a solid example, it would add a decent amount of cost for clustering

tarikceric: not worth it, reverted

eric-laurello (approved):

PR #811, opened 2025-03-03
An 5755/gold ez nft sales

tarikceric:
[Diagram](https://www.figma.com/board/FEgtAnUWAqC8CSztLy97OA/Solana-EZ-NFT-Sales?node-id=0-1&p=f&t=u1Udnr2eIlbwqxcY-0)

Create `ez_nft_sales` table
- Follow EVM standard column naming where applicable
- clean up logic from old `fact` table via new intermediate silver table for deprecated models + setting variables for existing models

FR on 2xl
```
20:51:16  1 of 2 OK created sql view model silver.nft_sales_legacy_combined_view ......... [SUCCESS 1 in 4.15s]
20:52:20  2 of 2 OK created sql incremental model nft.ez_nft_sales ....................... [SUCCESS 1 in 63.67s]
```

Inc on med
`1 of 1 OK created sql incremental model nft.ez_nft_sales ....................... [SUCCESS 4 in 12.06s]
`

Test:
```
21:50:55  Finished running 52 data tests, 7 project hooks in 0 hours 1 minutes and 2.54 seconds (62.54s).
21:50:56  Completed successfully
21:50:56  Done. PASS=52 WARN=0 ERROR=0 SKIP=0 TOTAL=52
```

Query to confirm counts (new table has 1 extra hadeswap sale that for some reason is not in fact):

```
with a as (select count(*) as ct_new_ez, program_id from solana_dev.nft.ez_nft_sales
where block_timestamp::date <= '2025-03-01'
group by program_id),

b as (select count(*) as ct_legacy_fact, program_id from solana.nft.fact_nft_sales
where block_timestamp::date <= '2025-03-01'
and succeeded
group by program_id)

select a.program_id, ct_new_ez, ct_legacy_fact from a
left join b
on a.program_id = b.program_id
where ct_new_ez <> ct_legacy_fact
;
```

### DIFF
--- a/models/gold/nft/nft__ez_nft_sales.sql
+++ b/models/gold/nft/nft__ez_nft_sales.sql
@@ -176,7 +176,7 @@ UNION ALL
     {% endfor %}
 )
 
-select 
+SELECT
     a.marketplace,
     a.marketplace_version,
     a.block_timestamp,
@@ -186,14 +186,14 @@ select
     a.index,
     a.inner_index,
     a.program_id,
-    a.purchaser as buyer_address,
-    a.seller as seller_address,
+    a.purchaser AS buyer_address,
+    a.seller AS seller_address,
     a.mint,
     b.nft_name,
-    a.sales_amount as price,
+    a.sales_amount AS price,
     a.currency_address,
-    c.symbol as currency_symbol,
-    (c.price * a.sales_amount) as price_usd,
+    c.symbol AS currency_symbol,
+    (c.price * a.sales_amount) AS price_usd,
     a.tree_authority,
     a.merkle_tree,
     a.leaf_index,
@@ -208,14 +208,12 @@ select
     a.ez_nft_sales_id,
     a.inserted_timestamp,
     a.modified_timestamp
-from base_nft_sales a
-left join 
+FROM
+    base_nft_sales a
+LEFT JOIN
     {{ ref('nft__dim_nft_metadata') }} b
-on a.mint = b.mint
-
-    LEFT JOIN {{ ref('price__ez_prices_hourly') }} c
+    ON a.mint = b.mint
+LEFT JOIN
+    {{ ref('price__ez_prices_hourly') }} c
     ON currency_address = c.token_address
-    AND DATE_TRUNC(
-        'hour',
-        a.block_timestamp
-    ) = c.hour
+    AND DATE_TRUNC('hour', a.block_timestamp) = c.hour
--- a/models/gold/nft/nft__ez_nft_sales.sql
+++ b/models/gold/nft/nft__ez_nft_sales.sql
@@ -215,5 +215,10 @@ LEFT JOIN
     ON a.mint = b.mint
 LEFT JOIN
     {{ ref('price__ez_prices_hourly') }} c
-    ON currency_address = c.token_address
+    ON (
+        CASE 
+            WHEN a.currency_address = 'So11111111111111111111111111111111111111111' THEN 'So11111111111111111111111111111111111111112'
+            ELSE a.currency_address
+        END
+    ) = c.token_address
     AND DATE_TRUNC('hour', a.block_timestamp) = c.hour
--- a/models/gold/nft/nft__ez_nft_sales.sql
+++ b/models/gold/nft/nft__ez_nft_sales.sql
@@ -200,7 +200,7 @@ SELECT
     b.nft_name,
     a.sales_amount AS price,
     a.currency_address,
-    c.symbol AS currency_symbol,
+    d.symbol AS currency_symbol,
     (c.price * a.sales_amount) AS price_usd,
     a.tree_authority,
     a.merkle_tree,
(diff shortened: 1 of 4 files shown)


desmond-hui (commented):
Is the goal here to eventually deprecate `fact_nft_sales`? The diagram makes it look like that it should be an upstream of `ez_nft_sales` instead but I don't think that is the case anymore?

If the above is true, can you compare the records in this table vs. the existing fact table to make sure there are no diffs in sales records?

tarikceric:
> Is the goal here to eventually deprecate `fact_nft_sales`? The diagram makes it look like that it should be an upstream of `ez_nft_sales` instead but I don't think that is the case anymore?
> 
> If the above is true, can you compare the records in this table vs. the existing fact table to make sure there are no diffs in sales records?

Updated the diagram to show it is not upstream of fact_nft_sales. Also, included a query in comments on ct comparison

models/gold/nft/nft__ez_nft_sales.sql
```diff
@@ -0,0 +1,232 @@
...
+)
+
+SELECT
+    a.marketplace,
+    a.marketplace_version,
+    a.block_timestamp,
+    a.block_id,
+    a.tx_id,
+    a.succeeded,
+    a.index,
```
desmond-hui, line 194: this has NULLs

```
select *
from solana_dev.nft.ez_nft_sales
where index is null;
```

models/gold/nft/nft__ez_nft_sales.sql
```diff
@@ -0,0 +1,232 @@
...
+    a.inner_index,
+    a.program_id,
+    a.purchaser AS buyer_address,
+    a.seller AS seller_address,
+    a.mint,
+    b.nft_name,
+    a.sales_amount AS price,
+    a.currency_address,
+    c.symbol AS currency_symbol,
```
desmond-hui, line 203: We don't have SOL prices before `2021-12-16 04:00:00.000` so alot of these become NULL. Also should we be getting symbol from the prices table vs. the dim metadata table?

models/gold/nft/nft__ez_nft_sales.yml
```diff
@@ -0,0 +1,167 @@
...
+        tests:
+          - dbt_expectations.expect_column_to_exist
+          - not_null: *recent_date_filter
+      - name: CURRENCY_SYMBOL
+        description: "Symbol of token used to pay for the NFT"
+        tests:
+          - dbt_expectations.expect_column_to_exist
+          - not_null: *recent_date_filter
+      - name: PRICE_USD
+        description: "Amount paid for NFT in USD"
```
desmond-hui, line 90: We should document that USD prices are not available before `2021-12-16 04:00:00.000` because we dont have pricing data before that

models/gold/nft/nft__ez_nft_sales.sql
```diff
@@ -0,0 +1,232 @@
...
+FROM
+    base_nft_sales a
+LEFT JOIN
+    {{ ref('nft__dim_nft_metadata') }} b
+    ON a.mint = b.mint
+LEFT JOIN
+    {{ ref('price__ez_prices_hourly') }} c
+    ON (
+        CASE 
+            WHEN a.currency_address = 'So11111111111111111111111111111111111111111' THEN 'So11111111111111111111111111111111111111112'
```
desmond-hui, line 228: Should use `SOL_MINT` here

models/gold/nft/nft__ez_nft_sales.sql
```diff
@@ -0,0 +1,232 @@
...
+    a.leaf_index,
+    a.is_compressed,
+    b.nft_collection_name,
+    b.collection_id,
+    b.creators,
+    b.authority,
+    b.metadata,
+    b.image_url,
+    b.metadata_uri,
+    a.ez_nft_sales_id,
```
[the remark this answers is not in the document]
tarikceric: The NULLs were mainly from older models that we never backfilled the _id for -- manually updated them and query returns 0.

The dupes were from duplicate entries in the dim_nft_metadata which i manually cleaned up

models/gold/nft/nft__ez_nft_sales.sql
tarikceric: Legacy models that were missing indexes -- manually updated the affected silver tables and query returns null

models/gold/nft/nft__ez_nft_sales.sql
tarikceric: The `currency_symbol` is related to the price and wouldn't be in the `dim_nft_metadata` table, unless I'm misunderstanding?

models/gold/nft/nft__ez_nft_sales.yml
tarikceric: Updated the table desc to note that

models/gold/nft/nft__ez_nft_sales.sql
desmond-hui: I meant `solana.price.dim_asset_metadata`

tarikceric: I feel like its fine to use the prices table instead of the metadata. It would be `null` for those older SOL prices, but so would the price anyway and not worth the extra join imo. The currency should all be fairly popular tokens that we would have prices for. Plus there's dupe addresses in the solana.price.dim_asset_metadata that would need some extra logic:

 select * from solana.price.dim_asset_metadata
where token_address in (select distinct(currency_address) from solana_dev.nft.ez_nft_sales);

desmond-hui (approved):

models/gold/nft/nft__ez_nft_sales.sql
tarikceric: updated

models/gold/nft/nft__ez_nft_sales.sql
tarikceric: joined to ez_asset_metadata

PR #813, opened 2025-03-05
AN-5835-Wormhole-Fix

gregoriustanleyy:
Fixing dupes on Wormhole transfers. Issue caused by outbound cte not having enough filters that it includes burns on Meteora too. Caught from daily test.


tarikceric:
after deleting the last few days of data and running the incremental i noticed its filtering out some bridge txs

  select * from solana.silver.bridge_wormhole_transfers
  where tx_id not in (select tx_id from solana_dev.silver.bridge_wormhole_transfers where block_timestamp::Date between '2025-03-01' and '2025-03-03')
  and block_timestamp::Date between '2025-03-01' and '2025-03-03';
  
 Ie tx 53cNbFQXtgVrfRTXsgMpttgQeKtAvn6JSRcPaKkS2kFReWnNkXSwot3gYcSEwuV93AWUkHuTnXKy2PRqQHwi7PYm -- i used wormholescan to doublecheck https://wormholescan.io/#/tx/53cNbFQXtgVrfRTXsgMpttgQeKtAvn6JSRcPaKkS2kFReWnNkXSwot3gYcSEwuV93AWUkHuTnXKy2PRqQHwi7PYm?network=Mainnet

models/silver/bridges/silver__bridge_wormhole_transfers.yml
```diff
@@ -27,50 +27,54 @@ models:
...
+          - not_null: *recent_date_filter
       - name: MINT
         description:  "{{ doc('mint') }}"
         tests: 
-          - not_null
+          - not_null: *recent_date_filter
       - name: _INSERTED_TIMESTAMP
         description: "{{ doc('tx_id') }}"
         tests: 
-          - not_null
```
tarikceric, line 76: that filter doesnt work on _inserted_timestamp as it will never return anything so we can remove it

PR #817, opened 2025-03-07
extra core schedule to fill gap

tarikceric:
- create `dbt_run_incremental_intermediate` workflow, to add extra run of core models as there is a gap between runs of `dbt_run_incremental_intermediate` and `dbt_run_incremental_non_core` every 4th hour
- hardcode concurrency


.github/workflows/dbt_run_incremental_intermediate.yml
```diff
@@ -0,0 +1,46 @@
+name: dbt_run_scheduled
+run-name: dbt_run_scheduled
```
eric-laurello, line 2: gotta update the name here

eric-laurello (approved):

PR #845, opened 2025-05-22
exclude stake snapshot; add alerting

tarikceric:
Tests now passing after filtering out snapshot stake accounts job (which runs on a custom trigger):

```
1 of 2 PASS dbt_expectations_expect_column_values_to_be_in_set_github_actions__current_task_status_PIPELINE_ACTIVE__True  [PASS in 15.73s]
2 of 2 PASS dbt_expectations_expect_column_values_to_be_in_set_github_actions__current_task_status_SUCCESSES__2  [PASS in 15.73s]
```

### DIFF
--- a/.github/workflows/dbt_test_tasks.yml
+++ b/.github/workflows/dbt_test_tasks.yml
@@ -25,3 +25,10 @@ jobs:
   called_workflow_template:
     uses: FlipsideCrypto/analytics-workflow-templates/.github/workflows/dbt_test_tasks.yml@AN-4374/upgrade-dbt-1.7
     secrets: inherit
+
+  notify-failure:
+    needs: [run_dbt_jobs]
+    if: failure()
+    uses: ./.github/workflows/slack_notify.yml
+    secrets:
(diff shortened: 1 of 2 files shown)


.github/workflows/dbt_test_tasks.yml
```diff
@@ -24,4 +24,11 @@ concurrency:
 jobs:
   called_workflow_template:
     uses: FlipsideCrypto/analytics-workflow-templates/.github/workflows/dbt_test_tasks.yml@AN-4374/upgrade-dbt-1.7
-    secrets: inherit
\ No newline at end of file
+    secrets: inherit
+
+  notify-failure:
+    needs: [run_dbt_jobs]
```
eric-laurello, line 30: I think we may need to change this job name?

tarikceric: ahh forgot that, updated

eric-laurello: not sure if it will work, but that's my best guess

eric-laurello (approved):

